PR #16, opened 2017-11-16
Increase default rate limits

kevinbader:
The same defaults are used by Amazon for its API Gateway product. The
previous limits were probably way too conservative.


mmacai:
Looks fine, merging.

PR #19, opened 2017-11-28
Describe how configuration should be handles (ADR 0004)

### LINKED ISSUE #11: Supersede ADR #0003, describe usage of Config macro
The [current ADR about the preferred handling of configuration](https://github.com/Accenture/reactive-interaction-gateway/blob/cf45bdfc1a2c6b917d20fddfea697fe68ff0d8f5/doc/architecture/decisions/0003-for-config-prefer-prefix-over-nesting-and-don-t-hide-defaults-in-code.md) is no longer valid. The preferred way of dealing with configuration is by using the Config macro module.

kevinbader:
Closes #11.


mmacai:
Looks fine I can agree with that.

PR #25, opened 2017-12-14
Maintain a changelog file

mmacai:
## Changes

- added `CHANGELOG.md` and related ADR
- filled `CHANGELOG.md` with all recent changes
- `mix docs` should work and generate new content

## Notes

- once we'll have umbrella setup we can add `###` headers and differentiate between direct changes in given sub application
e.g.
```
### Proxy

- Changed
  - ...

- Fixed
  - ...

### Monitoring

- Changed
  - ...
```

Decided to use PR numbers/links in notes. **Downside** is that you have to either guess (look at most recent PR and increment that by 1) your PR number or do one more commit after PR submit. But I think it's **better** to be redirected to PR and see code immediately rather than be redirected to issue and go through possibly long page to find out proper PR change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,30 @@
+# Changelog
+
+[Unreleased]
+
+- Changed
+  - [Config] Increase default rate limits - [#16](https://github.com/Accenture/reactive-interaction-gateway/pull/16)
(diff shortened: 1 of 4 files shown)


kevinbader:
Thanks, fully agree to the changes.

Regarding finding out the PR number - I'd just create the PR with [WIP] in the title and if the number was incorrent just amend and force-push the branch 🕺

PR #29, opened 2018-01-09
Kubernetes-Native

### LINKED ISSUE #28: Kubernetes-Native
Aim is to make RIG a nice Kubernetes service that you "just deploy" and that can automatically find the other instances and wire itself up. 

AC: Able to be launched as Kubernetes-Service easily on "own" Kubernetes cluster as well as stock Kubernetes-as-a-Service (AWS / Azure / Google) offerings.

mmacai:
Closes #28 

- added discovery module using `peerage` library via DNS => should work with any DNS (so it's not k8s specific, peerage seems to use just Erlang API)
- added distributed mode deployment readme
- added k8s config example + readme

**Note:** Should be probably released with 2.0 since setting `NODE_HOST` and `NODE_COOKIE` env vars is now mandatory (thus is breaking change prior to 1.x). Was looking for a way how it could be set automatically to dummy values if not specified in env vars, but without success.

Test:

- Following readme for k8s deployment should do the job (using Minikube and build own image, since official image doesn't have this changes yet).

**Note:** Not sure if discovery module should be wrapped with our own supervisor, since it's using `peerage` that has it's own supervisor. Your opinion?

How does it work:

- If you set env var `DISCOVERY_TYPE` (right now just `dns`) RIG will start `peerage` module
- `peerage` will do poll every 5 seconds and using Erlang do lookup using DNS
- If finds new node outside of cluster, new node will be connected (or removed, if it's down)

- If you don't set env var `DISCOVERY_TYPE`, RIG will work same as it works right now


vm.args
```diff
@@ -0,0 +1,22 @@
...
+## Heartbeat management; auto-restarts VM if it dies or becomes unresponsive
+## (Disabled by default..use with caution!)
+##-heart
+
+## Enable kernel poll and a few async threads
+##+K true
+##+A 5
+
+## Increase number of concurrent ports/sockets
+##-env ERL_MAX_PORTS 4096
```
kevinbader: perhaps we should increase this by default

mix.exs
```diff
@@ -38,6 +38,7 @@ defmodule Rig.Mixfile do
   def application do
     [
       mod: {Rig.Application, []},
+      included_applications: [:peerage],
```
kevinbader: is there a specific reason for including this explicitly?

lib/rig/discovery/discovery.ex
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  defp use_discovery(nil) do
+    Logger.info("No discovery for cluster specified")
+  end
+
+  defp use_discovery("dns") do
+    conf = config()
+    Application.put_env(:peerage, :via, Peerage.Via.Dns)
+    Application.put_env(:peerage, :dns_name, conf.dns_name)
```
kevinbader: I like this! Using configuration for `:rig` to configure `:peerage` behind the curtain is a nice touch.

deployment/k8s/rig_dns.yml
```diff
@@ -0,0 +1,66 @@
...
+        image: accenture/reactive-interaction-gateway
+        ports:
+        - containerPort: 4000
+        env:
+        # Optional:
+        # Disable Kafka by default, remove/comment out if you want to test with Kafka,
+        # in that case also set KAFKA_HOSTS env var to point to Kafka
+        - name: KAFKA_ENABLED
+          value: "false"
+        # Let's have ave some testing API endpoints
```
kevinbader: should probably read "Let's have some test-endpoints"

vm.args
```diff
@@ -0,0 +1,22 @@
+## Name of the node
+-name rig@${NODE_HOST}
```
kevinbader: I couldn't find a way of setting `NODE_HOST` to a sensible IP address automatically in case the env var is not set :/

mix.exs
mmacai: Didn't find any other working way, how to include `peerage` library, but don't start it by default (so it's started only if needed).

vm.args
mmacai: Ye, me neither. Hopefully it's not that sucky, to set it to `localhost` if you don't care about distributed mode.

mmacai:
Sure, let's have this prepared and merge once umbrella is done.

mix.exs
kevinbader: 👍

deployment/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+Reactive Interaction Gateway (RIG) uses [Peerage library](https://github.com/mrluc/peerage) to do discovery in distributed mode (production Distillery release).
+
+**Note:** If you don't care about distributed mode and don't want to do discovery, follow just `General configuration` section and ignore rest of the text.
+
+## General configuration
+
+1. Node host - Every node in cluster needs to be discoverable by other nodes. For that Elixir/Erlang uses so called `long name` or `short name`. We are using `long name` which is formed in the following way `app_name@node_host`. `app_name` is in our case set to `rig`, but `node_host` is taken from environment variable `NODE_HOST`. This can be either IP or container alias or whatever that is routable in network by other nodes.
+
+1. Node cookie - Nodes in Erlang cluster use cookies as a form of authorization between them. Only nodes with the same cookie can communicate together. It can be either some plain string or ideally some generated hash, set it to `NODE_COOKIE` environment variable.
```
kevinbader: s/authorization/authentication/

and it's _always_ a plain string, even if it's a hash of something ;)

rel/config.exs
```diff
@@ -30,13 +30,13 @@ environment :dev do
   # dev mode.
   set dev_mode: true
   set include_erts: false
-  set cookie: :"mby8_WifcvH[I^0B0<)FSkUpadXF~hWXY`ct6=lY9kj5cotG*^KZAaz0u?RYY%%`"
+  set cookie: :"placeholder_cookie"
```
kevinbader: I haven't tried, but since the cookie is set in vm_args anyway, couldn't this line be removed?

mmacai: Unfortunately not :/, distillery needs cookie here and erlang needs cookie in vm_args, thus I put here just some dummy token, since it's taken from vm_args.

deployment/README.md
mmacai: True, I'll reword it.

PR #42, opened 2018-02-14
Smoke tests

### LINKED ISSUE #33: integration/smoke tests

mmacai:
- added Kafka consumer ready check utility function
- added smoke tests setup (with docker-compose & Elixir ExUnit)
- added smoke tests cases - proxy, kafka messaging, kafka ready check
- updated dependencies - **had some weird problems with versions while running app/tests**
- removed `.dockerignore` - unused => we are selectively picking files in Dockerfile anyway
- updated contribution guide

**Test:**

- `mix test` should work as before
- see contribution guide for steps how to run smoke tests (**they should be sufficient to understand how to run them, if not steps should be updated!**)

Closes #33


kevinbader:
> removed .dockerignore - unused => we are selectively picking files in Dockerfile anyway

do you think we should use `.dockerignore` instead?

mmacai:
> do you think we should use .dockerignore instead?

I read somewhere, that using `.dockerignore` is best practice and leads to less layers in images, so perhaps we can discuss it and leverage from it in new PR.

apps/rig_inbound_gateway/test/rig_web/presence/channel_test.exs
```diff
@@ -52,4 +61,39 @@ defmodule RigInboundGatewayWeb.Presence.ChannelTest do
...
+
+    expected_event = "message"
+    expected_payload = %{"user" => "testuser", "foo" => "bar"}
+    assert_broadcast ^expected_event, ^expected_payload, 3000
+
+    leave sock
+  end
+
+  @tag :smoke
+  test "not subscribed user shouldn'\t receive message", %{conn: conn} do
```
kevinbader: tab character ... typo?

kevinbader:
Found one typo.. other than that it looks good to me

apps/rig_inbound_gateway/test/rig_web/presence/channel_test.exs
mmacai: ye, that's a typo ... used to do automatic escaping of special characters from JS

kevinbader:
Instructions worked like charm, tests seem to work.

apps/rig_outbound_gateway/lib/rig_outbound_gateway/kafka/group_subscriber.ex
```diff
@@ -114,13 +114,14 @@ defmodule RigOutboundGateway.Kafka.GroupSubscriber do
...
   """
+  def wait_for_consumer_ready(topic, partition, timeout \\ 10_000)
   @spec wait_for_consumer_ready(String.t(), integer, integer) :: {:error, String.t()}
   def wait_for_consumer_ready(topic, partition, timeout) when timeout <= 0 do
     {:error, "Consumer ready check for topic: #{topic} and partition: #{partition} timeouted."}
   end
 
   @spec wait_for_consumer_ready(String.t(), integer, integer) :: {:ok, String.t()}
-  def wait_for_consumer_ready(topic, partition, timeout \\ 10_000) do
+  def wait_for_consumer_ready(topic, partition, timeout) do
```
kevinbader: just a minor comment about the two specs: having one spec per function clause is optional and makes perfect sense if there something like "if I put a String.t in there, I get a String.t out, but if I put an atom in there, it'll be an atom" (you get the idea). However, in this case I feel it would be clearer (and less to write/read/maintain) to have just one combined spec, as I don't see any gain in expressing the same thing using two lines:
```elixir
@spec wait_for_consumer_ready(String.t(), integer, integer) :: {:ok | :error, String.t()}
```
What do you think?

PS: I think `integer` here should actually be `non_neg_integer()` in both cases

PR #43, opened 2018-02-15
Return helpful error on wrong content-type

kevinbader:


mmacai:
Cool, works as expected, let's merge it.

PR #50, opened 2018-02-28
Possibility to change LOG LEVEL with env var

### LINKED ISSUE #49: Possibility to change LOG LEVEL with env var
Make RIG's log level possible to configure via env var, otherwise app has to be recompiled to change it.

mmacai:
- added possibility to change logging level with env var
- it's a simple solution if we would like to support separate logger for umbrella app in future, we would probably need to add logger backend, for this this should suffice

**Test:**

- try to run test or app with and without `LOG_LEVEL` env var, should work as expected

Closes #49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
   - [Deploy] Kubernetes deployment configuration file - [#29](https://github.com/Accenture/reactive-interaction-gateway/pull/29)
   - [Misc] Smoke tests setup and test cases for API Proxy and Kafka + Phoenix messaging - [#42](https://github.com/Accenture/reactive-interaction-gateway/pull/42)
   - [Outbound] Kafka consumer ready check utility function - [#42](https://github.com/Accenture/reactive-interaction-gateway/pull/42)
-  - [Rig] Possibility to set logging level with env var `LOG_LEVEL` globally for RIG - [#49](https://github.com/Accenture/reactive-interaction-gateway/pull/49)
+  - [Rig] Possibility to set logging level with env var `LOG_LEVEL` globally for RIG - [#50](https://github.com/Accenture/reactive-interaction-gateway/pull/50)
 
 - Fixed
   - [Inbound] Make presence channel respect `JWT_USER_FIELD` setting (currently hardcoded to "username")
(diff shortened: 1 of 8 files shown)


apps/rig/lib/rig/application.ex
```diff
@@ -2,10 +2,15 @@ defmodule Rig.Application do
   @moduledoc false
 
   use Application
+  use Rig.Config, []
```
kevinbader: why not require it?
```elixir
use Rig.Config, [:log_level]
```

apps/rig/lib/rig/application.ex
```diff
@@ -2,10 +2,15 @@ defmodule Rig.Application do
...
 
   def start(_type, _args) do
     alias Supervisor.Spec
 
+    # Override application logging with environment variable
+    conf = config()
+    Logger.configure [{:level, conf.log_level}]
```
kevinbader: - not sure if we need to use `conf = ..` even in cases where we need it only once... I think I'd be okay with `config().log_level`, but ymmv
- formatting: parens for `configure` function call

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
   - [Deploy] Kubernetes deployment configuration file - [#29](https://github.com/Accenture/reactive-interaction-gateway/pull/29)
   - [Misc] Smoke tests setup and test cases for API Proxy and Kafka + Phoenix messaging - [#42](https://github.com/Accenture/reactive-interaction-gateway/pull/42)
   - [Outbound] Kafka consumer ready check utility function - [#42](https://github.com/Accenture/reactive-interaction-gateway/pull/42)
+  - [Rig] Possibility to set logging level with env var `LOG_LEVEL` globally for RIG - [#50](https://github.com/Accenture/reactive-interaction-gateway/pull/50)
```
kevinbader: - "globally for RIG" could be omitted
- I wanted to discuss this already - wouldn't it make more sense to refer to the issue instead of the PR?
  - there should be a Github issue for each changelog entry
  - the PR should contain closes/fixes <issue>, which will link the PR to the issue
  - you don't have to adapt the changelog entry if the actual PR number is different from what you've expected

apps/rig/lib/rig/application.ex
mmacai: was experimenting with something and didn't put it back

CHANGELOG.md
mmacai: This was my reasoning in changelog PR:

>Decided to use PR numbers/links in notes. Downside is that you have to either guess (look at most recent PR and increment that by 1) your PR number or do one more commit after PR submit. But I think it's better to be redirected to PR and see code immediately rather than be redirected to issue and go through possibly long page to find out proper PR change.

I don't have problem with either approach.

PR #53, opened 2018-03-05
Different Docker deployments

### LINKED ISSUE #44: Distillery and Docker deployment overhaul (version upgrade)

mmacai:
- revamped deployment with docker & distillery
- switched Elixir/Erlang images to alpine
- bumped up distillery version
- added possibility to deploy kinesis version
- updated documentation
- updated dockerhub image build to produce images in following way:

```
reactive-interaction-gateway
reactive-interaction-gateway:aws
reactive-interaction-gateway:2.0.0
reactive-interaction-gateway:2.0.0-aws
```

**Test:**

- check dockerhub image deployment setup
- `docker build -t rig -f aws.dockerfile .`
- 
```
docker run --name rig \
-e NODE_HOST=127.0.0.1 \
-e NODE_COOKIE=magiccookie \
-e LOG_LEVEL=debug \
-e AWS_ACCESS_KEY_ID=your_key \
-e AWS_SECRET_ACCESS_KEY=your_key \
-e KINESIS_ENABLED=true \
-e KINESIS_STREAM=your_stream_name \
-e KINESIS_AWS_REGION=your_region \
-e KINESIS_APP_NAME=your_app_dynamodb_name \
-e KINESIS_OTP_JAR=/opt/sites/rig/kinesis-client/local-maven-repo/org/erlang/otp/jinterface/1.8.1/jinterface-1.8.1.jar \
rig
```
- RIG should connect, try to push some messages to Kinesis, RIG should consume them

- `docker build -t rig-basic .`
- RIG should work as it is now, just Kinesis won't work

Closes #44

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,23 @@ Currently we support two ways to deploy RIG: using Docker and using classical Er
 
 You can find list of all environment variables in [operator guide](guides/operator-guide.md).
 
-### Deployment using Docker
+### Deployment using Docker & Distillery
+
+We are using Distillery releases to be able to run production version of RIG correctly optimized and tuned. Distillery has multiple entrypoints that are used for release:
+
+* `rel/config.exs` - Includes configuration for different types of environment (e.g. dev, prod, staging, etc.). Here Distillery finds which umbrella applications to compile and to use custom `vm.args` config file.
+* `vm.args` - Specifies options for Erlang VM, including name of the node and it's cookie. **Note:** cookie is also set in `rel/config.exs`, however it's just a dummy value and real value for it comes from this file (reason behind is that Distillery needs cookie set in `rel/config.exs` and Erlang in `vm.args` file, skipping either of those leads to error in build/production phase).
+* `config/` - Configuration for RIG itself (most of it can be set by environment variables on runtime, see [operator guide](guides/operator-guide.md)).
+* `apps/` - RIG's source folders.
+
+All these files are compiled together with an external dependencies and released as a tarball.
+
+By running built Docker image in container, release is executed in a foreground.
+
+There are multiple Dockerfiles that can be build:
+
+* `Dockerfile` - Has everything tha RIG offers, no exclusions.
+* `slim.dockerfile` - Excludes Java, thus AWS Kinesis is not possible to use.
 
 ```bash
 # Build the image:
@@ -84,21 +100,20 @@ export HOST_IP="$(ifconfig en0 inet | grep 'inet ' | awk '{ print $2 }')"
 docker run \
 --name rig \
 -p 4000:4000 \
+-e KAFKA_ENABLED=true \
 -e KAFKA_HOSTS="${HOST_IP}:9092" \
 -e NODE_HOST=localhost \
 -e NODE_COOKIE=magiccookie \
 rig
 
 # Check that the proxy api is indeed available on port 4000 (returns an empty list by default):
-curl localhost:4000/apis
+curl localhost:4000/v1/apis
 ```
 
 ### Deployment using Erlang Releases
 Using Erlang releases (instead of Docker containers) allows for hot code reloading. At the same
 time, you have to take care of cross-compilation and
 [the hiccups of code hot-loading](http://learnyousomeerlang.com/relups#the-hiccups-of-appups-and-relups).
-
-TODO describe Distillery builds and perhaps hot code reloading.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -88,8 +88,8 @@ By running built Docker image in container, release is executed in a foreground.
 
 There are multiple Dockerfiles that can be build:
 
-* `Dockerfile` - Has everything tha RIG offers, no exclusions.
-* `slim.dockerfile` - Excludes Java, thus AWS Kinesis is not possible to use.
+* `Dockerfile` - Basic version, excludes Java, thus AWS Kinesis is not possible to use.
+* `aws.dockerfile` - Includes Java which is required for AWS Kinesis.
 
 ```bash
 # Build the image:
--- a/README.md
+++ b/README.md
@@ -84,12 +84,12 @@ We are using Distillery releases to be able to run production version of RIG cor
 
 All these files are compiled together with an external dependencies and released as a tarball.
 
-By running built Docker image in container, release is executed in a foreground.
+When running built Docker image in a container, release is run in a foreground.
 
 There are multiple Dockerfiles that can be build:
 
-* `Dockerfile` - Basic version, excludes Java, thus AWS Kinesis is not possible to use.
-* `aws.dockerfile` - Includes Java which is required for AWS Kinesis.
+* `Dockerfile` - Standard image, suitable for on-premise use.
+* `aws.dockerfile` - AWS-flavored version that let's RIG integrate with AWS services like Amazon Kinesis.
 
 ```bash
 # Build the image:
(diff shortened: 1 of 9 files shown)


aws.dockerfile
```diff
@@ -0,0 +1,99 @@
...
+COPY vm.args /opt/sites/rig/
+
+# Copy necessary files for dependencies
+COPY mix.exs /opt/sites/rig/
+COPY mix.lock /opt/sites/rig/
+COPY apps/rig/mix.exs /opt/sites/rig/apps/rig/
+COPY apps/rig_api/mix.exs /opt/sites/rig/apps/rig_api/
+COPY apps/rig_auth/mix.exs /opt/sites/rig/apps/rig_auth/
+COPY apps/rig_inbound_gateway/mix.exs /opt/sites/rig/apps/rig_inbound_gateway/
+COPY apps/rig_outbound_gateway/mix.exs /opt/sites/rig/apps/rig_outbound_gateway/
```
kevinbader: I thought we've settled on using dockerignore instead of this mount of COPY statements...

aws.dockerfile
```diff
@@ -0,0 +1,99 @@
...
+ENV REPLACE_OS_VARS=true
+
+# Install Java
+RUN { \
+		echo '#!/bin/sh'; \
+		echo 'set -e'; \
+		echo; \
+		echo 'dirname "$(dirname "$(readlink -f "$(which javac || which java)")")"'; \
+	} > /usr/local/bin/docker-java-home \
+	&& chmod +x /usr/local/bin/docker-java-home
```
kevinbader: why is [`apk add openjdk8-jre`](https://pkgs.alpinelinux.org/package/edge/community/x86/openjdk8-jre) not enough?

aws.dockerfile
```diff
@@ -0,0 +1,99 @@
+FROM maven:3-jdk-8-alpine as java-build
+
+COPY kinesis-client /opt/sites/rig/kinesis-client
```
kevinbader: here you also copy the jinterface jar that is used during compilation and also later at runtime:
```
KINESIS_OTP_JAR=/opt/sites/rig/kinesis-client/local-maven-repo/org/erlang/otp/jinterface/1.8.1/jinterface-1.8.1.jar
```

However, I'd prefer having the runtime compiled with Java support and using the runtime jinterface jar for both compilation and runtime. In that case, the KINESIS_OTP_JAR variable doesn't have to be set, because RIG is expecting the jinterface jar at the right place (as defined by the spec).

aws.dockerfile
```diff
@@ -0,0 +1,99 @@
...
+
+# Compile AWS Kinesis Java application
+RUN mvn package
+
+FROM elixir:1.6-alpine as elixir-build
+
+# Install Elixir & Erlang environment dependencies
+RUN mix local.hex --force
+RUN mix local.rebar --force
+RUN apk update && apk add make \
```
kevinbader: `--no-cache` instead of update

aws.dockerfile
```diff
@@ -0,0 +1,99 @@
...
+
+COPY apps/rig_outbound_gateway/config /opt/sites/rig/apps/rig_outbound_gateway/config
+COPY apps/rig_outbound_gateway/lib /opt/sites/rig/apps/rig_outbound_gateway/lib
+
+# Compile and release application production code
+RUN mix release
+
+FROM erlang:20-alpine
+
+RUN apk update && apk add bash
```
kevinbader: `--no-cache` instead of `update`

README.md
```diff
@@ -73,7 +73,23 @@ Currently we support two ways to deploy RIG: using Docker and using classical Er
...
+We are using Distillery releases to be able to run production version of RIG correctly optimized and tuned. Distillery has multiple entrypoints that are used for release:
+
+* `rel/config.exs` - Includes configuration for different types of environment (e.g. dev, prod, staging, etc.). Here Distillery finds which umbrella applications to compile and to use custom `vm.args` config file.
+* `vm.args` - Specifies options for Erlang VM, including name of the node and it's cookie. **Note:** cookie is also set in `rel/config.exs`, however it's just a dummy value and real value for it comes from this file (reason behind is that Distillery needs cookie set in `rel/config.exs` and Erlang in `vm.args` file, skipping either of those leads to error in build/production phase).
+* `config/` - Configuration for RIG itself (most of it can be set by environment variables on runtime, see [operator guide](guides/operator-guide.md)).
+* `apps/` - RIG's source folders.
+
+All these files are compiled together with an external dependencies and released as a tarball.
+
+By running built Docker image in container, release is executed in a foreground.
```
kevinbader: s/By/When/
s/container/a container/
s/executed in a/run in/

README.md
```diff
@@ -73,7 +73,23 @@ Currently we support two ways to deploy RIG: using Docker and using classical Er
...
+
+There are multiple Dockerfiles that can be build:
+
+* `Dockerfile` - Basic version, excludes Java, thus AWS Kinesis is not possible to use.
```
kevinbader: I wouldn't specify this here. Actually, we don't know if we'll ever need Java in the main image (I really hope we don't). What about
```
Dockerfile -- Standard image, suitable for on-premise use.
aws.dockerfile -- AWS-flavored version that let's RIG integrate with AWS services like Amazon Kinesis.
```

kinesis-client/.dockerignore
```diff
@@ -0,0 +1,8 @@
+.settings
+target
+.classpath
+.editorconfig
+.gitignore
+.project
+dependency-reduced-pom.xml
+README.md
```
kevinbader: the local-maven-repo should be here too, see related comment above

aws.dockerfile
mmacai: based on official Java Dockerfile https://github.com/docker-library/openjdk/blob/9a0822673dffd3e5ba66f18a8547aa60faed6d08/8-jdk/alpine/Dockerfile might try if this will suffice

aws.dockerfile
mmacai: based on official Java Dockerfile https://github.com/docker-library/openjdk/blob/9a0822673dffd3e5ba66f18a8547aa60faed6d08/8-jdk/alpine/Dockerfile might try if this will suffice

aws.dockerfile
mmacai: tbh, read some articles and `.dockerignore` is not always the best approach, but I'll think about it ...

mmacai: Problem with general `.dockerignore` here is that e.g. if we put tests there, then dockerfile for smoke tests will exclude them as well, so no tests to run. There is a long discussion about it here https://github.com/moby/moby/issues/12886, but no solution to this day. I would not heavily rely on `.dockerignore` till it's solved by guys from Docker.

aws.dockerfile
kevinbader: We've settled on not compiling it for now, but instead using the checked-in version of JInterface (OtpErlang.jar). Because of that, I feel that we should provide the right default for the path inside the image; right now, when setting `KINESIS_ENABLED=1`, you see the following error:

```plain
"Could not find OtpErlang.jar for JInterface 1.8.1 at /opt/sites/rig/lib/jinterface-1.8.1/priv/OtpErlang.jar. Does your Erlang distribution come with Java support enabled?"
```

I propose that we simply use `ENV` in the dockerfile to point to the right path, so the user doesn't have to set this later on:

```Dockerfile
ENV KINESIS_OTP_JAR=/opt/sites/rig/kinesis-client/local-maven-repo/org/erlang/otp/jinterface/1.8.1/jinterface-1.8.1.jar
```

PR #56, opened 2018-03-21
Easier getting-started guide + JWT script

kevinbader:


kevinbader:
The deployment documentation still needs some love, we'll fix that in another PR

PR #61, opened 2018-04-04
Handle content-type correctly (api, /messages)

kevinbader:
- Allow MIME type options like ';charset=utf-8' for json content-type
- Allow application/x-www-form-urlencoded and multipart encodings
- Return "415 Unsupported Media Type" for unknown content-types


mmacai:
Works, but perhaps you should add few small tests (for `application/json;charset=utf-8`, `application/x-www-form-urlencoded`, some unsupported media type) to be sure that it will work also in the future (we may decide at some point to add/update something in this controller).

PR #62, opened 2018-04-04
fixed overriding of user defined proxy query auth values

lofim:
There is an issue with default `auth` values overriding the user defined ones when the `auth_type` is set to "jwt".

Default `auth` values are not part of the `default_api_values` before we even merge with `api`.
JWT required `header_name` and `use_header` attributes ares are apply retroactively in a way that does not overrides the user defined query attributes.

It possible to test this with the testing proxy json and setting the query attributes to `true` and `token` for example. In master those will be overridden to `false` and `""`.


kevinbader:
Thanks for the contribution, Martin. Looking through the code I've noticed that I would've expected a slightly different behavior in case the auth type is "jwt" and "use_query" is set:

```diff
diff --git a/apps/rig_inbound_gateway/test/rig/proxy_test.exs b/apps/rig_inbound_gateway/test/rig/proxy_test.exs
index 3306ae5..041a0dc 100644
--- a/apps/rig_inbound_gateway/test/rig/proxy_test.exs
+++ b/apps/rig_inbound_gateway/test/rig/proxy_test.exs
@@ -180,6 +180,8 @@ defmodule RigInboundGateway.ProxyTest do
 
       assert api["auth"]["use_query"]
       assert api["auth"]["query_name"] == "token"
+      refute api["auth"]["use_header"]
+      refute api["auth"]["header_name"] == nil
 
       assert ctx.tracker |> Stubr.called_thrice?(:track)
       assert ctx.tracker |> Stubr.called_twice?(:find_by_node)
```

This fails, so I've wondered whether this is intended behavior (would be okay, too), or the header setting is supposed to get cleared in case the query parameter is set. In any case, please add the relevant assertions.

lofim:
I've added the assertions, but to give you a bit of a context.

The idea here was to not override the user defined _auth query_ values with the _jwt auth type_ option. And have them **both** present in the resulting configuration as opposed to just having one or another.

I have this use case where I need to restrict access to a static file (an image), however I am unable (or lacking the knowledge) to send the token as part of the header since I access the image via a link. The token is then part of the URL. 

The service that handles the image over relies on token being in the header for every other endpoint. That's why I need both to be present in the RIG config.

I am not sure if this is the correct way of dealing with this but looking at other parts of the code 
[api_proxy/router.ex$L106](https://github.com/Accenture/reactive-interaction-gateway/blob/master/apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/router.ex#L106):
```
tokens = Enum.concat(Auth.pick_query_token(conn, api), Auth.pick_header_token(conn, api))
```
 I got the impression it should be possible to have both at the same time.

Therefore the PR.

kevinbader:
@lofim aaah now I understand.. I must admit I wasn't aware of this use case, thanks for the explanation 🙌

PR #70, opened 2018-05-06
JS dependency vulnerable updates

mmacai:
Warnings were caused by `hoek` dependency.

For `Hapi` server it's solved in this commit.

For `create-react-app` frontend it's stated here (so I would wait until it's fixed on their side):
https://github.com/facebook/create-react-app/issues/4374


Urnamma:
Approved.

PR #76, opened 2018-05-22
Headers transformation

mmacai:
- Fixed regex matching for proxy - previously `/abc/def` would match against `/abc/def/1`
- Fixed response headers before response send - Elixir/Plug needs lower cased headers => led to duplicate headers and wrong responses
- Added request headers transformations via `add_headers`, in future we can add `remove_headers`, etc.

- Unit tests and smoke tests should cover everything

Steps to test:
- create new API at `POST` `localhost:4010/v1/apis` with
```
{
  "id": "new-service2",
  "name": "new-service2",
  "auth_type": "jwt",
  "auth": {
    "use_header": true,
    "header_name": "Authorization",
    "use_query": false,
    "query_name": ""
  },
  "versioned": false,
  "version_data": {
    "default": {
    	"transform_request_headers": {
          "add_headers": {
            "a": "b",
            "c": "d"
          }
        },
		"endpoints": [
			{
				"id": "get-auth-register",
				"path": "/auth/register",
				"method": "GET",
				"not_secured": true,
				"transform_request_headers": true
			}
		]
    }
  },
  "proxy": {
    "use_env": true,
    "target_url": "IS_HOST",
    "port": 6666
  }
}
```
- Send request to `GET` `localhost:4000/auth/register` with header `a` => proxy should overwrite it and add new header `b`
- Omitting `"transform_request_headers": true` in endpoints will ignore header transformations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
   - [Rig] Possibility to set logging level with env var `LOG_LEVEL` - [#49](https://github.com/Accenture/reactive-interaction-gateway/pull/49)
   - [Deploy] Variations of Dockerfiles - basic version and AWS version - [#44](https://github.com/Accenture/reactive-interaction-gateway/pull/44)
   - [Deploy] Helm deployment chart - [#59](https://github.com/Accenture/reactive-interaction-gateway/pull/59)
+  - [Inbound] Proxy is now able to do request header transformations - [#76](https://github.com/Accenture/reactive-interaction-gateway/pull/76)
 
 - Fixed
   - [Inbound] Make presence channel respect `JWT_USER_FIELD` setting (currently hardcoded to "username")
@@ -32,6 +33,8 @@
   - [API] Set proper environment variable for Phoenix server `API_PORT` - [#38](https://github.com/Accenture/reactive-interaction-gateway/pull/38)
   - [Examples] Channels example fixed to be compatible with version 2.0.0 [#40](https://github.com/Accenture/reactive-interaction-gateway/pull/40)
   - [Inbound] User defined query auth values are no longer overridden by `JWT` auth type
+  - [Inbound] More strict regex match for routes in proxy - [#76](https://github.com/Accenture/reactive-interaction-gateway/pull/76)
+  - [Inbound] Downcased response headers to avoid duplicates in proxy - [#76](https://github.com/Accenture/reactive-interaction-gateway/pull/76)
 
 - Deprecated
 


kevinbader:
```
  "versioned": false,
  "version_data": {
    "default": {
```

I don't really get this saying theres no versioning, but then configuring "version_data" ... I know it's done the same way in Tyk, and they probably had a reason for doing so, but it just looks very awkward to me 😅

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/serializer.ex
```diff
@@ -33,18 +35,35 @@ defmodule RigInboundGateway.ApiProxy.Serializer do
...
+  # Add new headers and update existing ones
+  @spec add_headers(headers_list, headers_list) :: headers_list
+  def add_headers(new_headers, old_headers) do
+    old_headers
+    |> Enum.filter(fn({old_key, _}) ->
+      new_headers
+      |> Enum.find(nil, fn({new_key, _}) -> old_key == new_key end)
+      |> is_nil
+    end)
+    |> Enum.concat(new_headers)
```
kevinbader: hm maybe I get this wrong, but it seems this
1. takes all old headers that don't occur in `new_headers` and
2. add the new headers to that list.
Would that be easier to read if written this way:
```elixir
Map.merge(Map.new(old_headers), Map.new(new_headers))
|> Enum.to_list()
```

apps/rig_inbound_gateway/test/rig/api_proxy/router_test.exs
```diff
@@ -225,6 +225,36 @@ defmodule RigInboundGateway.ApiProxy.RouterTest do
...
+    end
+
+    request = build_conn(:post, "/myapi/no-transform-headers") |> put_req_header("host", "original")
+    conn = call(Router, request)
+
+    assert get_req_header(conn, "host") == ["original"]
+    assert get_req_header(conn, "john") == []
+    assert conn.status == 200
+    assert conn.resp_body =~ "{\"status\":\"ok\"}"
+  end
```
kevinbader: it'd be nice to have a test around how it behaves with `versioned` set or unset

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/router.ex
```diff
@@ -91,25 +91,47 @@ defmodule RigInboundGateway.ApiProxy.Router do
...
     case Auth.any_token_valid?(tokens) do
-      true -> forward_request(endpoint, api, conn)
+      true -> transform_req_headers(endpoint, api, conn)
       false -> send_resp(conn, 401, Serializer.encode_error_message("Missing or invalid token"))
     end
   end
 
+  # Transform request headers
+  @spec transform_req_headers(Proxy.endpoint(), Proxy.api_definition(), %Plug.Conn{}) ::
+          %Plug.Conn{}
```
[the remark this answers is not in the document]
mmacai: Didn't know about `Plug.Conn.t`, but I think it makes sense since Strings are defined the same way.

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/serializer.ex
mmacai: Definitely, totally overlooked it.

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/serializer.ex
```diff
@@ -33,18 +35,35 @@ defmodule RigInboundGateway.ApiProxy.Serializer do
...
-    conn
-    |> Map.get(:resp_headers)
-    |> Enum.find({}, fn(headers_tuple) ->
+  @spec header_value?(headers_list, String.t, String.t) :: boolean
+  def header_value?(headers, key, value) do
+    headers
+    |> Enum.find({}, fn({header_key, _}) -> header_key == key end)
+    |> Tuple.to_list
+    |> Enum.member?(value)
+  end
```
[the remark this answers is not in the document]
mmacai: Nice catch, added test as well for it

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/router.ex
```diff
@@ -91,25 +91,47 @@ defmodule RigInboundGateway.ApiProxy.Router do
...
+  defp transform_req_headers(
+         %{"transform_request_headers" => true} = endpoint,
+         %{"versioned" => false} = api,
```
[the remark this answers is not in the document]
kevinbader: but doesn't `%{"transform_request_headers" => true} = endpoint` say that the transformation should be enabled _on this specific endpoint_?

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/router.ex
```diff
@@ -171,24 +193,26 @@ defmodule RigInboundGateway.ApiProxy.Router do
...
+  @spec send_response({:ok, Plug.Conn.t, map}) :: Plug.Conn.t
   defp send_response({:ok, conn, %{headers: headers, status_code: status_code, body: body}}) do
-    conn = %{conn | resp_headers: headers}
-    if Serializer.header_value?(conn, "transfer-encoding", "chunked") do
-      send_chunked_response(conn, headers, status_code, body)
+    downcased_headers = headers |> Serializer.downcase_headers
+    conn = %{conn | resp_headers: downcased_headers}
+
+    if Serializer.header_value?(downcased_headers, "transfer-encoding", "chunked") do
+      send_chunked_response(conn, downcased_headers, status_code, body)
```
kevinbader: why are the headers passed here again - `conn` should already include them at this point

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/serializer.ex
```diff
@@ -7,6 +7,8 @@ defmodule RigInboundGateway.ApiProxy.Serializer do
   alias Plug.Conn.Query
   alias RigInboundGateway.Proxy
 
+  @typep headers :: [{String.t, String.t}, ...]
```
kevinbader: What if old_headers is empty? I mean, sending a request without any headers would look more like [{String.t, String.t}]

apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/router.ex
mmacai: *facepalm*

Issue #55, opened 2018-03-16
Endpoint for disconnect_channel_session should either make use of the user parameter or be models to not require it at all


lofim:
The RG API router contains a route for deleting user sessions (disconnecting user based on his JTI) and it looks like this:
```
delete "/:user/sessions/:jti", ChannelsController, :disconnect_channel_session
```

However the `:user` parameter is not used and I can happily send whatever string I want.
```
Request URL: http://localhost:3000/rg/users/undefined/sessions/8bf39303-1a10-4cf8-9765-b00f82ef258b
Request Method: DELETE
Status Code: 204 No Content
```

Notice the `undefined` part of the URL.

The handler should either check for existence of the user and that it contains the session we want to delete (as the REST semantics make it appear).
Or the route should be modeled in a different way not containing the user parameter at all.

PR #78, opened 2018-05-25
Removed user id parameter from terminate session api endpoint

lofim:
Addressing #55.

As discussed in the issue linked above I've changed the terminate session endpoint to not contain user id in the path.

This is a pretty simple change. Hopefully I didn't forget anything.

The old endpoint has been removed as I am not sure how we are handling endpoint deprecation.
This is a breaking change after all, supposed land in v2.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v2.0.0-dev
 
 - Changed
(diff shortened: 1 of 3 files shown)


lofim:
Sure, go ahead.

PR #80, opened 2018-06-08
Support Kafka SSL encryption and authentication

### LINKED ISSUE #79: Support Kafka SSL encryption and authentication
Relevant resources:

- [Confluent Encryption and Authentication with SSL](https://docs.confluent.io/current/kafka/authentication_ssl.html)
- Brod example client cert config at [runtime](https://github.com/klarna/brod/blob/master/test/brod_client_SUITE.erl#L309) and in [sys.config](https://github.com/klarna/brod/blob/master/sys.config.example#L29)
- [Brod example setup for SASL gssapi (kerberos) authentication](https://github.com/klarna/brod/wiki/SASL-gssapi-(kerberos)-authentication)

kevinbader:
Closes #79 

## Test

Verify that the Dockerized smoke tests run through:

```bash
$ cd smoke_tests
$ ./run_smoke_tests.sh
...
╓─────[ All tests passed. ]─────╖
```

Verify that the dev setup still works as well:

```bash
$ KAFKA_ENABLED=1 mix phx.server
...
15:28:47.626 [info]  Group member (rig-consumer-group,coor=#PID<0.793.0>,cb=#PID<0.781.0>,generation=10):
assignments received:
  rig:
    partition=0 begin_offset=8
```

Other things to look at: changelog entry, documentation/code.


chadmott:
I cannot speak for the _code_ changes but I've tested a build of this branch with our implementation of RIG and it works. 

One suggestion, the confluent docs suggest encrypting the private key with a password (https://docs.confluent.io/current/kafka/authentication_ssl.html) so to get it to work, i had to decrypt the key the other clients were using. Would it be possible to implement a password mechanism similar to this?

kevinbader:
@chadmott thanks for the review. Supporting password-protected private keys is a good idea indeed. In theory, it shouldn't be hard to implement, because Brod simply passes on the SSL options to Erlang's `ssl` module. Doesn't seem to work that way though, at least not the way I do it. If you're interested, you can follow the issue here: https://stackoverflow.com/questions/50813212/erlang-ssl-password-option-for-private-key

mmacai:
Ok, seems that there is a problem with cert generator (when using passwords). Used certificates from this generator https://github.com/edenhill/librdkafka/wiki/Using-SSL-with-librdkafka and even password protected certificates worked.

kevinbader:
After investigating some more I found out that the Erlang `ssl` implementation doesn't play well with the private key created by `openssl req -newkey`. In the openssl documentation I couldn't find the exact algorithm used by `-newkey`, but creating the private key beforehand using `openssl genrsa -des3` did the trick, so I think the default algorithm is something not supported by Erlang's `ssl`.

apps/rig_outbound_gateway/lib/rig_outbound_gateway/kafka/sup.ex
```diff
@@ -40,10 +40,22 @@ defmodule RigOutboundGateway.Kafka.Sup do
...
+    end
+
+    client_conf =
+      [
+        endpoints: conf.brokers,
+        auto_start_producers: true,
+        default_producer_config: [],
+        ssl: conf.ssl
+      ]
+      |> add_sasl_option.()
```
mmacai: Do you think we will at some point pass argument here? Or is there any other reason why it's not directly passed as ssl config is?

smoke_tests/run_smoke_tests.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo ""
+}
+
+section_header "Building docker image: ${DOCKER_IMAGE}"
+docker build -f ../smoke_tests.dockerfile -t "${DOCKER_IMAGE}" ..
+
+section_header "Starting Kafka and backend-simulation services"
+$COMPOSE up -d kafka rest-api
+
+section_header "Running tests, plaintext Kafka connection"
```
mmacai: Thumbs up for testing both cases!

apps/rig_outbound_gateway/lib/rig_outbound_gateway/kafka/sup.ex
```diff
@@ -40,10 +40,22 @@ defmodule RigOutboundGateway.Kafka.Sup do
...
+    add_sasl_option = fn opts ->
+      if conf.sasl, do: Keyword.put(opts, :sasl, conf.sasl), else: opts
+    end
+
+    client_conf =
+      [
+        endpoints: conf.brokers,
+        auto_start_producers: true,
+        default_producer_config: [],
+        ssl: conf.ssl
```
mmacai: I guess this covers also SSL setup for producer, right?

smoke_tests/run_smoke_tests.sh
```diff
@@ -0,0 +1,33 @@
...
+  -e KAFKA_SSL_ENABLED=0 \
+  -e KAFKA_HOSTS=kafka:9292 \
+  rig
+
+section_header "Running tests, encrypted Kafka connection"
+$COMPOSE run --rm \
+  -e KAFKA_ENABLED=1 \
+  -e KAFKA_SSL_ENABLED=1 \
+  -e KAFKA_HOSTS=kafka:9393 \
+  -e KAFKA_SSL_KEYFILE_PASS=abcdef \
```
mmacai: Shouldn't this be `abcdefgh`? Or Am I missing something?

apps/rig_outbound_gateway/lib/rig_outbound_gateway/kafka/sup.ex
kevinbader: `ssl` is normally a keyword list, but when it's disabled it's set to false. The `sasl` argument works differently: if disabled, it must not be present. I couldn't figure out a nice way to write this inline or with a pipe, so I've opted for the function..

smoke_tests/run_smoke_tests.sh
kevinbader: _Ideally_ we would not rerun _all_ the tests, but yeah, should be alright for now

smoke_tests/run_smoke_tests.sh
kevinbader: Indeed, I've just discovered an error in the script that generates the certificates, which caused the private to be stored _unencrypted_. This is why the wrong password still worked, and I'm glad that you've actually catched that. Turns out that `openssl genpkey` simply ignores the `-pass` parameter if there is no algorithm selected. When I wrote this, I foolishly assumed that either an algorithm would be selected by default, or that there would be a warning in case I'd have to select one. `openssl` being `openssl`, I guess.

apps/rig_outbound_gateway/lib/rig_outbound_gateway/kafka/sup.ex
kevinbader: I've added a [test for this](https://github.com/Accenture/reactive-interaction-gateway/blob/79-kafka-ssl/apps/rig/test/kafka_test.exs#L22):
```
Rig.KafkaTest
  * test publishing using an actual connection (66.2ms)
```

apps/rig_outbound_gateway/lib/rig_outbound_gateway/kafka/sup.ex
mmacai: makes sense

kevinbader:
I've added a small but really nice hook script to the release:
```bash
$ docker run -it --rm rig:ssl
NODE_HOST not set, defaults to 127.0.0.1. Consider setting NODE_HOST to the machine's hostname or IP, as seen by others in the network.
NODE_COOKIE not set; randomly generated to hBaHutUsD2g9Pi19Ww35ptzjtOp4iuma
Reactive Interaction Gateway 2.0.0-beta.1 [rig@127.0.0.1, ERTS 10.0.4, OTP 21]
```
No more staring at no output and wondering whether RIG already runs or the command didn't work ^_^

mmacai:
Since `NODE_HOST` and `NODE_COOKIE` are now auto-generated we should also update README - line `58` and below.

PR #87, opened 2018-08-03
Fix helm chart

### LINKED ISSUE #69: Helm chart issues
The helm chart config has 2 issues.

The first issue is related to port configuration.
In `deployment.yaml` template there are 2 entries for ports (api and inbound) with their names specified as `api-port` and `inbound-port`. This is fine but the config is missing the environment variables (`API_PORT` and `INBOUND_PORT`) that would tell RIG on what port to run.

The second issue is that the `values.yaml` file declares DNS name as `reactive-interaction-gateway-service-headless.default.svc.cluster.local` but the service is created without the `-service` suffix. So the correct name would be `reactive-interaction-gateway-headless.default.svc.cluster.local`.

I'll create a PR for this when I get some time.

mmacai:
- addressed second issue in #69 - updated value for DNS in helm chart
- first issue with port is understandable, but I would keep this chart example simple to be 1:1 with kubernetes config (to be able to compare it)

closes #69

### DIFF
--- a/deployment/helm/reactive-interaction-gateway/templates/deployment.yaml
+++ b/deployment/helm/reactive-interaction-gateway/templates/deployment.yaml
@@ -32,6 +32,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: API_PORT
+              value: {{ .Values.apiPort | quote }}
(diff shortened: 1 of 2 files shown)


lofim:
Awesome, works jus fine. Let's merge it.

PR #111, opened 2018-12-06
Feature/109 health endpoint

vanillebear:
This is to close issue 109. Additionally including minor documentation improvements describing the API for session_Blacklist.

Short summary:
- New route /health
- New Controller for endpoint
- Unit test for endpoint

- minor documentation improvement (Independent from code fix)

### DIFF
--- a/apps/rig_api/lib/rig_api/router.ex
+++ b/apps/rig_api/lib/rig_api/router.ex
@@ -32,4 +32,11 @@ defmodule RigApi.Router do
       delete("/:id", ApisController, :deactivate_api)
     end
   end
+
+  scope "/health", RigApi do
+    pipe_through(:api)
+    scope "/" do
+      get("/", HealthController, :check_health)
+    end
+  end
 end
--- a/apps/rig_api/test/rig_api/controllers/health_controller_test.exs
+++ b/apps/rig_api/test/rig_api/controllers/health_controller_test.exs
@@ -11,4 +11,3 @@ defmodule RigApi.HealthControllerTest do
       end
     end
   end
-  
--- a/apps/rig_api/test/rig_api/controllers/health_controller_test.exs
+++ b/apps/rig_api/test/rig_api/controllers/health_controller_test.exs
@@ -1,0 +1,14 @@
+defmodule RigApi.HealthControllerTest do
+    @moduledoc false
+    require Logger
+    use ExUnit.Case, async: true
+    use RigApi.ConnCase
+  
+    describe "GET /health" do
+      test "should return OK as text response" do
+        conn = build_conn() |> get("/health")
+        assert conn.resp_body =~ "OK"
+      end
+    end
+  end
+  
--- a/apps/rig_api/lib/rig_api/controllers/health_controller.ex
+++ b/apps/rig_api/lib/rig_api/controllers/health_controller.ex
@@ -8,4 +8,3 @@ defmodule RigApi.HealthController do
       text(conn, "OK")
     end
   end
-  
--- a/apps/rig_api/lib/rig_api/controllers/health_controller.ex
+++ b/apps/rig_api/lib/rig_api/controllers/health_controller.ex
@@ -1,0 +1,11 @@
+defmodule RigApi.HealthController do
+    require Logger
+  
+    use RigApi, :controller
+  
+    @doc "Default response for health status"
+    def check_health(conn, _params) do
+      text(conn, "OK")
+    end
+  end
+  
--- a/guides/configuration.md
+++ b/guides/configuration.md
@@ -118,6 +118,17 @@ cluster, so all nodes share the same configuration without having to sync anythi
 ##### Delete API
 `DELETE /v1/apis/:api_id`
 
+##### Register session to blacklist
+`POST /v1/session-blacklist/`
+```json
+{
+	"validityInSeconds": "10",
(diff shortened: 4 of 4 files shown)


vanillebear:
@kevinbader there seems to be an issue with the build of the Website due to some credential issues:

remote: Invalid username or password.
fatal: Authentication failed for 'https://kevinbader@github.com/Accenture/reactive-interaction-gateway.git/'

vanillebear:
For reference here the test log:
`
vanillebear@Jakob-Surface-Go:/mnt/c/projekte/reactive-interaction-gateway$ mix test test/rig_api/controllers/health_controller_test.exs
warning: path "apps/rig_metrics" is a directory but it has no mix.exs. Mix won't consider this directory as part of your umbrella application. Please add a "mix.exs" or set the ":apps" key in your umbrella configuration with all relevant apps names as atoms
==> rig_kafka
Paths given to mix test did not match any directory/file: test/rig_api/controllers/health_controller_test.exs
==> rig
Paths given to mix test did not match any directory/file: test/rig_api/controllers/health_controller_test.exs
==> rig_auth
Paths given to mix test did not match any directory/file: test/rig_api/controllers/health_controller_test.exs
==> rig_inbound_gateway
Paths given to mix test did not match any directory/file: test/rig_api/controllers/health_controller_test.exs
==> rig_outbound_gateway
Paths given to mix test did not match any directory/file: test/rig_api/controllers/health_controller_test.exs
==> rig_api
Excluding tags: [smoke: true]
.
**Finished in 0.3 seconds
1 test, 0 failures**
Randomized with seed 427982
==> rig_tests
Paths given to mix test did not match any directory/file: test/rig_api/controllers/health_controller_test.exs
`

apps/rig_api/lib/rig_api/router.ex
```diff
@@ -32,4 +32,11 @@ defmodule RigApi.Router do
       delete("/:id", ApisController, :deactivate_api)
     end
   end
+
+  scope "/health", RigApi do
+    pipe_through(:api)
+    scope "/" do
```
kevinbader: Would it also work without that inner `scope`? For example, see line 11.

apps/rig_api/test/rig_api/controllers/health_controller_test.exs
```diff
@@ -0,0 +1,13 @@
+defmodule RigApi.HealthControllerTest do
+  @moduledoc false
+  require Logger
+  use ExUnit.Case, async: true
+  use RigApi.ConnCase
+
+  describe "GET /health" do
+    test "should return OK as text response" do
+      conn = build_conn() |> get("/health")
+      assert conn.resp_body =~ "OK"
```
kevinbader: You could use `==` here instead of `=~` since you expect the full body to match. If you'd like to assert that the response is actually a "text response", you could also assert the mime type to be `text/plain`.

kevinbader:
> @kevinbader there seems to be an issue with the build of the Website due to some credential issues

thanks - should be fixed in master now

apps/rig_api/lib/rig_api/router.ex
vanillebear: fixed - see latest commit

apps/rig_api/test/rig_api/controllers/health_controller_test.exs
vanillebear: fixed - see latest commit

apps/rig_api/test/rig_api/controllers/health_controller_test.exs
```diff
@@ -0,0 +1,15 @@
...
+  require Logger
+  use ExUnit.Case, async: true
+  use RigApi.ConnCase
+
+  describe "GET /health" do
+    test "should return OK as text response" do
+      conn = build_conn() |> get("/health")
+      content_type = get_resp_header(conn, "content-type")
+      assert conn.resp_body == "OK"
+      assert content_type == ["text/plain; charset=utf-8"]
```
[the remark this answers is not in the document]
vanillebear: Hi! The suggested change doesn't exactly work like proposed as the ignore isn't allowed as it cannot be used in expressions… additionally as a list is returned from `get_resp_header(..)` -->`Plug.Conn.Utils.content_type() `throws an error…

Following code would work:
 `assert {:ok, "text", "plain", %{}} == conn |> get_resp_header("content-type") |> hd() |> String.split(";") |> hd() |> Plug.Conn.Utils.content_type()`

Is this still fine with you? Then I'll commit that change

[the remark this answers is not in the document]
kevinbader: I get the feeling this is something we might need in other tests too, so we could put it into a dedicated function and potentially move it into another module later on. Please consider this:

```suggestion
      assert "text/plain" == resp_content_type(conn)
```

with

```suggestion
  # The response content-type with parameters stripped (e.g. "text/plain").
  defp resp_content_type(conn)
      [{:ok, type, subtype, _params}] =
          conn
          |> get_resp_header("content-type")
          |> Enum.map(&Plug.Conn.Utils.content_type/1)

      "#{type}/#{subtype}"
  end
```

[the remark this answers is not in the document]
vanillebear: change tested & committed

PR #117, opened 2018-12-11
WIP: Feature/116 phoenix swagger

vanillebear:
**Attention: This PR is still WIP & isn't final yet.**

This is to initially show how this could look like & currently just one Controller was documented.
It is already including the auto-generated swagger.json

Please provide feedback if this seems to be useful for you. If not I'll close the issue

_screenshot of the swagger-ui which is exposed by a plug_
rig_swagger_ui


vanillebear:
Hello @kevinbader could you please check on this PR?

It's now documenting the Endpoints for sessionBlacklist, Apis, Health and message Controllers of the rig_api/rig_control

I'm not really sure what the "Channels" Controller does and how it is used. Anyways I have seen that lately 2 out of 3 Endpoints have been deprecated anyways.

kevinbader:
@vanillebear thanks again for the update,

> It's now documenting the Endpoints for sessionBlacklist, Apis, Health and message Controllers of the rig_api/rig_control
> 
> I'm not really sure what the "Channels" Controller does and how it is used. Anyways I have seen that lately 2 out of 3 Endpoints have been deprecated anyways.

You're right, the channels controller has mostly been deprecated as the API it provides has turned out to be less than ideal.

I've checked out your branch and tried merging with master, but I got this:

```
[warn]  Failed to generate priv/static/rig_api_swagger.json: Failed to load module: Elixir.RigApi.Router: nofile
```

Perhaps you already know what's wrong here. Also, please take a look at [the merge commit on my (temporary) branch](https://github.com/Accenture/reactive-interaction-gateway/tree/kevin-116-phoenix-swagger), where I've also edited the Changelog entry a bit.

README.md
```diff
@@ -160,6 +160,11 @@ source.addEventListener("greeting", function (e) {
 - JWT signature verification for APIs as a simple authentication check.
 - Session blacklist with immediate session invalidation.
 
+### API Documentation
+The API documentation for RIG can be found when running the RIG under the endpoint /swagger-ui e.g. when running on locally: [localhost:4010/swagger-ui](localhost:4010/swagger-ui)
```
kevinbader: RIG exposes its API documentation on its API endpoint under `/swagger-ui` (e.g. http://localhost:4010/swagger-ui).

apps/rig_api/lib/rig_api/controllers/health_controller.ex
```diff
@@ -2,6 +2,16 @@ defmodule RigApi.HealthController do
...
   use RigApi, :controller
+  use PhoenixSwagger
+
+  # Swagger documentation for endpoint GET /health
+  swagger_path :check_health do
+    get("/health")
+    summary("Provides the RIG health status")
+    description("Provides the RIG health status")
+
+    response(200, "Ok")
```
kevinbader: "Ok" vs. "OK" in the actual response

apps/rig_api/lib/rig_api/router.ex
```diff
@@ -37,4 +37,29 @@ defmodule RigApi.Router do
...
+
+  scope "/swagger-ui" do
+    forward("/", PhoenixSwagger.Plug.SwaggerUI,
+      otp_app: :rig_api,
+      swagger_file: "rig_api_swagger.json"
+    )
+  end
+
+  def swagger_info do
+    %{rig: rig_version} = versions()
```
kevinbader: Please replace this with `Application.spec(:rig, :vsn)`. The "version" file might not be present in a release. The reason for the `Code.eval_file` thing in the Mix files is that it is currently not possible to globally set the project version for an umbrella project, but that's just Mix and it's not meant to be used in production code.

vanillebear: Unfortunately `Application.spec(:rig, :vsn)` responds `nil`… `RigApi.Mixfile.project[:version]` would work?!

vanillebear:
@kevinbader updated the docs as suggested.

For the "Version" I used another method as `Application.spec` provided `nil`... Hope that's fine..

Regarding your error… Unfortunately I haven't seen this yet and for me it's working. 
You're using `mix phx.server` to run the RIG, right?
Also the ` Elixir.RigApi.Router` hasn't been renamed or sth, so I actually don't know the issue :/

apps/rig_api/lib/rig_api/router.ex
kevinbader: not sure we're trying the same thing here.. For me, this works:

```bash
$ iex -S mix
iex(1)> Application.spec(:rig, :vsn)
'2.0.0-beta.2'
```
```bash
$ cd apps/rig_api
$ iex -S mix
iex(1)> Application.spec(:rig, :vsn)
'2.0.0-beta.2'
```
```bash
$ mix release
$ _build/dev/rel/rig/bin/rig console
iex(rig@127.0.0.1)1> Application.spec(:rig, :vsn)
'2.0.0-beta.2'
```

The main problem with your proposed solution is that `RigApi.Mixfile` is simply not present in a release; in fact, any Mix related functionality is basically only available in dev and test.

vanillebear: I did test this in the exact same way as you did. Same result, I'm getting '2.0.0-beta.2' as a response.

Still, when I use this in router.exs, like below, I'm getting a nil.
```
  def swagger_info do
    %{
      info: %{
        version: Application.spec(:rig, :vsn),
        title: "RIG Control API",
        description: "This is the description for the RIG Control API exposed on Port 4010 by default or API_PORT within the config-file.
        It manages the Proxy APIs or user connections for RIGs proxy."
      }
    }
  end
```

kevinbader:
> For the "Version" I used another method as `Application.spec` provided `nil`... Hope that's fine..

I've added a comment to the discussion above..

> Regarding your error… Unfortunately I haven't seen this yet and for me it's working.
> You're using `mix phx.server` to run the RIG, right?
> Also the ` Elixir.RigApi.Router` hasn't been renamed or sth, so I actually don't know the issue :/
> Have you merged the mix-file? In [PhoenixSwagger Issue 176](https://github.com/xerions/phoenix_swagger/issues/176) they described that phoenix_swagger was missing in the compilers

Found the cause: the `:phoenix_swagger` compiler relies on artifacts compiled by the Elixir compiler. Therefore, I suggest to modify the ordering:

```diff
diff --git a/apps/rig_api/mix.exs b/apps/rig_api/mix.exs
index 10e923c..6a08f80 100644
--- a/apps/rig_api/mix.exs
+++ b/apps/rig_api/mix.exs
@@ -13,7 +13,7 @@ defmodule RigApi.Mixfile do
       lockfile: "../../mix.lock",
       elixir: elixir_version,
       elixirc_paths: elixirc_paths(Mix.env()),
-      compilers: [:phoenix, :gettext, :phoenix_swagger] ++ Mix.compilers(),
+      compilers: [:phoenix, :gettext] ++ Mix.compilers() ++ [:phoenix_swagger],
       start_permanent: Mix.env() == :prod,
       aliases: aliases(),
       deps: deps(),
```

Test: `mix clean; rm -rf _build; rm ./apps/rig_api/priv/static/rig_api_swagger.json; mix compile` previously failed to generate `rig_api_swagger.json` but with the change it no longer does.

vanillebear:
> Found the cause: the :phoenix_swagger compiler relies on artifacts compiled by the Elixir compiler. Therefore, I suggest to modify the ordering:
Thank you for this… Didn't know that the order of the Compilers matter :/ This was already updated in the latest commit

> I've added a comment to the discussion above
OK - Continuing discussion there

apps/rig_api/lib/rig_api/router.ex
vanillebear: Hi @kevinbader I think the reason that `Application.spec(:rig, :vsn) `returns Nil in this case is, that actually the app isn't started yet.
This is because Phoenix_swagger uses it's own mix Task `phx.swagger.generate` prior starting the actual app.

I'm not sure if this part of code is ever in use in a release setup?! Shouldn't the swagger.json only change during development & is therefore part of a commit?
Hence, Maybe it's valid to use `RigApi.Mixfile` as we only need it in dev

What's your thought's on this?

vanillebear: Hi @kevinbader:

Finally got it. with `Application.ensure_all_started` the Generation of the swagger.json is forced to be executed after start. :) This works both when executing `mix compile` & `mix phx.server`

**Downside**: Application is temporary started during `mix compile`

Here's how I implemented it now & included in the latest commit :)

```
  def swagger_info do
    Application.ensure_all_started(:rig_api, :temporary) 
    %{
      info: %{
        version: Application.spec(:rig_api,:vsn) |> to_string,
        title: "RIG Control API",
        description: "This is the description for the RIG Control API exposed on Port 4010 by default or API_PORT within the config-file.
        It manages the Proxy APIs or user connections for RIGs proxy."
      }
    }
  end
```

mix.exs
```diff
@@ -27,7 +28,7 @@ defmodule Rig.Umbrella.Mixfile do
 
   defp versions do
     {map, []} = Code.eval_file("version", ".")
-    map
+    map
```
kevinbader: added whitespace.. If you're not doing this already, consider using format-on-save (i.e. `mix format`) with your editor - it fixes changes like this automatically

apps/rig_api/lib/rig_api/router.ex
```diff
@@ -37,4 +37,23 @@ defmodule RigApi.Router do
...
+      otp_app: :rig_api,
+      swagger_file: "rig_api_swagger.json"
+    )
+  end
+
+  def swagger_info do
+    Application.ensure_all_started(:rig_api, :temporary) 
+    %{
+      info: %{
+        version: Application.spec(:rig_api,:vsn) |> to_string,
```
kevinbader: whitespace (comma)... just as a comment, I've never seen `to_string` explicitly used anywhere; I think the idiomatic way to do this would be

```elixir
"#{Application.spec(:rig_api, :vsn)}"
```

apps/rig_api/lib/rig_api/router.ex
kevinbader: nice idea! I'm not really convinced that starting an application is the way to go just to obtain the release version though. Will look into this tomorrow

kevinbader: @vanillebear 

seems like the compile task and the mix task do the same thing:

> PhoenixSwagger provides a mix compiler and mix task phx.swagger.generate for the swagger-ui json file generation that contains swagger specification that describes API of the phoenix application.

=> I'd remove the alias. The compile task should ensure that the file is current also when building a release.

> I'm not sure if this part of code is ever in use in a release setup?!

You're right, I got the wrong - please use what you've had initially: `RigApi.Mixfile.project[:version]` is fine since this is never executed in a release context. Sorry for the confusion.

> Shouldn't the swagger.json only change during development & is therefore part of a commit?

Isn't the JSON file a build artifact and as such out-of-scope for version control?

Issue #77, opened 2018-05-23
Support forwarding inbound HTTP request bodies as events to Kafka & Kinesis


kevinbader:
Right now, HTTP requests are forwarded to backend services. Additionally, other "targets" should be available, like publishing the request to a Kafka or Kinesis topic.

#### Benefits

- If a service is down at the time the request is sent, the request is not lost but can be processed as soon as the service is back up.
- Frontends don't see 504 Gateway Timeout errors, since RIG immediately responds. Assuming the processing service notifies the frontends about picking up the request, the UI just has another state to deal with (e.g., submitted - processing - processed).

LeDominik:
Current plan: Release as Beta-feature in 2.0. We already know that this will be superceded with proper async support in #97, with this initial feature then being the fire & forget mode.

mmacai:
Done in #123

PR #123, opened 2018-12-13
Proxy events stream fix & Bearer authorization header

### LINKED ISSUE #82: Add support for token type in Authorization header
Since we support sending auth (bearer) token via Authorization header we should be compliant with what the rest of the industry is doing.

The [OAuth RFC 7.1](https://tools.ietf.org/html/rfc6749#section-7.1) describes token types.
The current implementation does not consider types at all and will fail to validate the token if the type is set to for example "Bearer".

So the proper form should look like this:
`Authorization : Bearer eyJ0eXAiOiJK`

Instead of this:
`Authorization : eyJ0eXAiOiJK`

AFAIK we only support Bearer token for now.
The question here is what should the gateway do in case it receives a different type.

Either way this should be configurable. We should be able to specify that we expect a type or not.
`AUTH_TOKEN_TYPE=Bearer` sounds reasonable to me.

I'm open to suggestions.

mmacai:
- fixed proxy requests to event streams - wrong formatting and pattern match for CORS
- added integration tests for kafka proxy publish
- changed jwt validation to expect proper official format - `Bearer token`

Closes #82 #77


CHANGELOG.md
```diff
@@ -9,15 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [Proxy] When forwarding requests, RIG related meta data (e.g. correlation ID) in CloudEvents is now put into an object under the top-level key "rig". Note that in terms of the current [CloudEvents 0.2](https://github.com/cloudevents/spec/blob/v0.2/spec.md) specification this makes "rig" an [extension](https://github.com/cloudevents/spec/blob/v0.2/primer.md#cloudevent-attribute-extensions). Also, all RIG related keys have been renamed from snake_case to camelCase.
 
 <!-- ### Deprecated -->
 
 <!-- ### Removed -->
 
-<!-- ### Fixed -->
+### Fixed
+
+- [Inbound] Wrong parsing and place in cloud events for API proxy publish to Kafka/Kinesis
```
kevinbader: Can you be more specific here? Like either describing how the parsing failed and what the place was and now is, or linking an issue.

CHANGELOG.md
```diff
@@ -9,15 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ### Added -->
 
-<!-- ### Changed -->
+### Changed
+
+- [Auth] JWT now expects proper format -- `Bearer ...`.
```
kevinbader: Can you link the part of the RFC you've implemented?

apps/rig_auth/lib/rig_auth/jwt/utils.ex
```diff
@@ -12,19 +12,27 @@ defmodule RigAuth.Jwt.Utils do
...
   @spec valid?(String.t()) :: boolean
-  def valid?(jwt) do
+  def valid?("Bearer " <> jwt) do
     jwt
     |> validate
     |> get_error == nil
   end
 
+  def valid?(jwt) do
+    %{error: "JWT=#{jwt} has incorrect form. Required form is: \"Bearer token\""}
```
kevinbader: Probably being picky here, but
- the spec doesn't call this "form"
- a non-bearer type token might not be supported by RIG but that doesn't necessarily mean it's an "incorrect" token
- the variable shouldn't be called `jwt` as it's most likely _not_ a JWT at this point

apps/rig_inbound_gateway/test/rig_web/presence/channel_test.exs
```diff
@@ -14,86 +14,96 @@ defmodule RigInboundGatewayWeb.Presence.ChannelTest do
   end
 
   test "a user connecting to her own topic works" do
-    assert {:ok, _response, sock} = subscribe_and_join_user(
-      "testuser",
-      [@customer_role],
-      "user:testuser"
-    )
-    leave sock
+    assert {:ok, _response, sock} =
```
kevinbader: seems the changes in this file are only around reformatting and commenting the smoke tests out. If you remove this file from the PR you would spare me one merge conflict ;)

apps/rig_tests/test/proxy/response_from/kafka_test.exs
```diff
@@ -115,4 +124,163 @@ defmodule RigTests.Proxy.ResponseFrom.KafkaTest do
     # ...but the client got the response sent to the Kafka topic:
     assert Jason.decode!(res_body)["message"] == Map.fetch!(async_response, "message")
   end
+
+  @tag :kafka
+  test "Given target is set to Kafka, the http OPTIONS request should handle CORS" do
```
kevinbader: as discussed today this should be moved out of the response_from module

CHANGELOG.md
```diff
@@ -9,15 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ### Added -->
 
-<!-- ### Changed -->
+### Changed
+
+- [Auth] JWT now expects proper format -- `Bearer ...` -- [rfc6749](https://tools.ietf.org/html/rfc6749#section-7.1).
```
kevinbader: what about

`- [Auth] No longer assumes the "Bearer" token type when no access token type is prepended in the Authorization header. Consequently, a client is expected to explicitly use "Bearer" for sending its JWT authorization token. More more details, see [RFC 6749](https://tools.ietf.org/html/rfc6749#section-7.1).`

CHANGELOG.md
```diff
@@ -9,15 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [Proxy] When forwarding requests, RIG related meta data (e.g. correlation ID) in CloudEvents is now put into an object under the top-level key "rig". Note that in terms of the current [CloudEvents 0.2](https://github.com/cloudevents/spec/blob/v0.2/spec.md) specification this makes "rig" an [extension](https://github.com/cloudevents/spec/blob/v0.2/primer.md#cloudevent-attribute-extensions). Also, all RIG related keys have been renamed from snake_case to camelCase.
 
 <!-- ### Deprecated -->
 
 <!-- ### Removed -->
 
-<!-- ### Fixed -->
+### Fixed
+
+- [Inbound] Wrong pattern match for CORS handling when sending proxy request to Kafka/Kinesis. Before -- `%{"method" => "OPTIONS"}`, after -- `%{method: "OPTIONS"}`.
```
kevinbader: Ah I didn't quite get what this was about. Do you think it makes sense to put implementation details into the Changelog? Assuming this file is targetted at end users, I guess something like this might suffice:

`- [Inbound] Fixed response to CORS related preflight request.`

apps/rig_auth/lib/rig_auth/jwt/utils.ex
```diff
@@ -12,19 +12,27 @@ defmodule RigAuth.Jwt.Utils do
...
   @spec valid?(String.t()) :: boolean
-  def valid?(jwt) do
+  def valid?("Bearer " <> jwt) do
     jwt
     |> validate
     |> get_error == nil
   end
 
+  def valid?(invalid_jwt) do
+    %{error: "JWT=#{invalid_jwt} is missing token type. Required format is: \"Bearer token\""}
```
kevinbader: I was thinking more about something like `invalid_access_token`, based on the spec.

apps/rig_tests/test/proxy/publish_to_event_stream/kafka_test.exs
```diff
@@ -0,0 +1,204 @@
...
+    assert res_body == "Accepted."
+
+    expected_msg =
+      Jason.encode!(%{
+        "source" => "rig-test",
+        "rig" => %{
+          "scheme" => "http",
+          "requestPath" => "/mock-proxy-http-response-publish-to-kafka-endpoint",
+          "reqHeaders" => [
+            ["user-agent", "hackney/1.14.0"],
```
kevinbader: I suppose this makes the test unnecessary brittle. The assertion should cover the `requestPath`, the `correlationID` and the `eventID`, and perhaps also the `method`. If `host` and `port` should be checked as well, then you should probably use `Confex` to fetch the actual values of `HOST`, and `@proxy_port` for the actual port, instead of assuming `localhost` and `4000`.

mmacai: I would rather check everything and be sure when something changes - not a fan of partial checking.

kevinbader: But do you really want to go fix tests every time hackney is upgraded? You got me curious.. why are you not a fan of partial checking?

mmacai: That's true, on the other hand if someone extends the message in code and do mistake there, but forget to update test we have error in app without knowing it.

kevinbader: I feel the simple question that needs to be asked for each value we test is: is there a reason, in the context of this test, for testing it? Given this example, the test assures that `Given target is set to Kafka, the http request should publish message to Kafka topic`, which describes the following assertion: the message sent to the proxy ends up at the Kafka topic. That's what this is about. Now, how can we check that the message ends up there? We may check the event ID or the whole message, depending on whether we have reason to believe the message could be altered in the process (e.g., the proxy doesn't use the body verbatim, so that's a valid concern). But that's about it - indeed, checking the added meta data could justify a second test, something like "When forwarding an event to Kafka, connection related meta data is added to the given CloudEvent.". In that test, of course, we'd assert the presence and correctness of the connection related meta data, but even in this test I doubt that the user-agent is of any relevance.

mmacai: Just noticed the test is anyway badly written, will do update.

PR #135, opened 2019-01-08
Support match-by-regex for proxy request path and path rewriting

### LINKED ISSUE #88: reverse proxy: support URL rewriting
Right now, `{id}` can be used as a placeholder for variable parts of an URL. There are issues with the current approach:

- The name inside the curly braces does not matter - it can't be referred to in the target URL.
- The format is non-standard and quite inflexible.

I think that if we want to support URL rewriting we should adopt industry standards, which basically means using regular expressions. Using regular expressions is admittedly a bit more complicated than the current approach might be, but it is very efficient (expressions can be pre-compiled) and flexible.

The implementation should enable the following scenario:

```plain
Given the request URL `/books/123/chapters/456`,
and the rule `/books/([^/]+)/chapters(?<chapter_id>/?|/.*)` => `http://chapter-service:8080/v1/chapters/chapter_id`,
then the request is forwarded to `http://chapter-service:8080/v1/chapters/456`.
```

Proposed [:re options](http://erlang.org/doc/man/re.html): `:no_auto_capture`, `:dupnames`, `:unicode`, `:anchored`, `:caseless`.

kevinbader:
Closes #88.

### DIFF
--- a/apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/api.ex
+++ b/apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/api.ex
@@ -10,7 +10,24 @@ defmodule RigInboundGateway.ApiProxy.Api do
           optional(:type) => String.t(),
           optional(:target) => String.t(),
           id: String.t(),
+          # Simple matching; curly braces may be used to ignore parts of the URI.
+          # Example:
+          #     /path/{to}/somewhere/{special} is matched by /path/1/somewhere/2
           path: String.t(),
+          # Matches against a regular expression.
+          # Note that JSON requires escaping the backslash character.
+          # Example:
+          #     /path/.+/somewhere/.+ is matched by /path/1/somewhere/2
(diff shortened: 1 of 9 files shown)


apps/rig_inbound_gateway/lib/rig_inbound_gateway/api_proxy/api.ex
```diff
@@ -10,7 +10,24 @@ defmodule RigInboundGateway.ApiProxy.Api do
...
           path: String.t(),
+          # Matches against a regular expression.
+          # Note that JSON requires escaping the backslash character.
+          # Example:
+          #     /path/.+/somewhere/.+ is matched by /path/1/somewhere/2
+          path_regex: String.t(),
+          # Used to rewrite the request path. When used with `path_regex`, capture
+          # groups can be referenced by number.
+          # Note that JSON requires escaping the backslash character.
+          # Example:
```
mmacai: +1 for examples

PR #160, opened 2019-02-06
Remove RateLimit implementation

### LINKED ISSUE #144: Revisit (rewrite or remove) rate limiting
Currently, the code around rate limiting is basically dead: inactive by default, with config values in the ops-guide marked as deprecated. We need to decide whether we need this feature. If we do, we need to check whether we can re-use existing code; otherwise, we should get rid of it.

jakobwgnr:
Removes RateLimit from SourceCode & Documentation

This is to close #144


jakobwgnr:
Hello @kevinbader or @mmacai  
Could someone of you please have a look on this PR?
Thanks in Advance!

jakobwgnr:
@kevinbader changelog was updated & committed

PR #165, opened 2019-02-13
81 avro

### LINKED ISSUE #81: Support Avro and the Confluent Schema Registry
Currently, RIG handles JSON only, but using Avro along with the Schema Registry seems to be the way to go when using Kafka.

Basically, we need something like this but with Brod instead of `KafkaEx`: https://github.com/sevenmind/kaufmann_ex

mmacai:
Closes #81 

Now prepared for review. See docs for more info.

~~This is still WIP work.~~

~~What's missing:~~

- ~~finish tests~~
- ~~docs~~
- ~~clean the code & simplification~~

~~Despite of not finished yet I would like to get some feedback for parts that can be for sure improved.~~


apps/rig_kafka/lib/rig_kafka/client.ex
```diff
@@ -28,11 +30,82 @@ defmodule RigKafka.Client do
...
+
+      deconstructed_headers =
+        headers
+        |> Enum.into(%{})
+        |> extract_headers()
+
+      decoded_body =
+        cond do
+          Map.get(deconstructed_headers, "contentType") == "avro/binary" ->
+            data = Jason.decode!(Serializer.decode_body(body, "avro"))
```
kevinbader: - if `Serializer.decode_body` fails, wouldn't it return an ok or error tuple?
- would it be reasonable to fall-back to another format in case decoding fails?

apps/rig_kafka/lib/rig_kafka/avro.ex
```diff
@@ -0,0 +1,115 @@
+defmodule RigKafka.Avro do
+  @moduledoc """
+  TODO
+  """
+
+  require Logger
+  use Memoize
+  use Rig.Config, [:schema_registry_host]
```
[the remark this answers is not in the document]
mmacai: Ye, I was looking for something like that, always irritates me when it's mixed.

apps/rig_kafka/lib/rig_kafka/avro.ex
```diff
@@ -0,0 +1,115 @@
...
+
+  @spec parse_schema(String.t()) :: {:ok, map()}
+  def parse_schema(subject) do
+    {:ok, %{"schema" => raw_schema, "id" => id}} = get(subject)
+    schema = :avro.decode_schema(raw_schema)
+    Logger.debug("Using Avro schema with name=#{subject}")
```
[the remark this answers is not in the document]
mmacai: Interesting, good to know.

apps/rig_kafka/lib/rig_kafka/client.ex
mmacai: Hmm, good point, I'll take a look.

apps/rig_kafka/lib/rig_kafka/avro.ex
```diff
@@ -0,0 +1,115 @@
...
+    {id, schema}
+  end
+
+  @spec parse_schema(integer) :: map()
+  def parse_schema_by_id(id) do
```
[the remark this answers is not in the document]
mmacai: Yes, `{:ok, map()}` is a mistake. Thing is the second function is returning also `ID` of schema, so if we overload a single function it will require some branching -- or you see some nice way how to have it as a 1 function and avoid if/else here and there?

Or we can use single call by ID, but to me it's more convenient to specify rather name than ID when producing event. What do you think?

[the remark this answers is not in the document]
kevinbader: I've had guards in mind. For example, you could use `is_number` for the ID case and default to string otherwise.

apps/rig_kafka/lib/rig_kafka/avro.ex
```diff
@@ -0,0 +1,104 @@
+defmodule RigKafka.Avro do
+  @moduledoc """
+  Module responsible for event serialization and deserialization. Manages also connection to Kafka Schema Registry and caching of used schemas.
```
kevinbader: wrap at 80?

integration_tests/kafka_tests/run.sh
```diff
@@ -18,19 +33,17 @@ source .env
 docker-compose up -d || exit 1
 
 while ! is_kafka_ready; do
-    echo "waiting for Kafka to start and create the topic.."
+    printf "waiting for Kafka to start and create the topics..\n"
```
kevinbader: just curious: why? :)

integration_tests/kafka_tests/run.sh
```diff
@@ -1,11 +1,26 @@
...
+    if [[ -z "$(docker-compose exec kafka bash -c 'kafka-topics --list --zookeeper zookeeper:2181')" ]]; then
         # no output - Kafka is not ready yet
         return 1
     else
         # Kafka responds!
+        section_header "Creating Kafka topics"
+        docker-compose exec kafka bash -c 'kafka-topics --create --topic rig_test --partitions 1 --replication-factor 1 --if-not-exists --zookeeper zookeeper:2181'
+        docker-compose exec kafka bash -c 'kafka-topics --create --topic rig-proxy-response --partitions 1 --replication-factor 1 --if-not-exists --zookeeper zookeeper:2181'
+        docker-compose exec kafka bash -c 'kafka-topics --create --topic rig_kafka_test_simple_topic --partitions 1 --replication-factor 1 --if-not-exists --zookeeper zookeeper:2181'
+        docker-compose exec kafka bash -c 'kafka-topics --create --topic rigAvro --partitions 1 --replication-factor 1 --if-not-exists --zookeeper zookeeper:2181'
```
kevinbader: 3 topic names, 3 casings: dashed, snake_case and camelCase ;)

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
+---
+id: event-serializing
+title: Event Serializing
+sidebar_label: Event Serializing
+---
+
+We support [Apache Avro](https://avro.apache.org/) and standard JSON (de)serialization of events for producer as well as for consumer. This is working together with [Cloud Events specification](https://github.com/cloudevents/spec) (versions `0.1` and `0.2`). To make both of these worlds work we are utilizing Kafka headers introduced in version `0.11`. As Cloud Events specify there are 2 modes how to make this combination work. Discussion about this topic can be followed on [Github](https://github.com/cloudevents/spec/pull/337/files).
```
kevinbader:
```suggestion
We support [Apache Avro](https://avro.apache.org/) and standard JSON (de)serialization of events for both Kafka consumers and producers. We aim at following the [CloudEvents specification](https://github.com/cloudevents/spec) as closely as possible (the related part of the spec is called _transport bindings_ and is, at the time of writing, still under development). For the Kafka transport binding, we utilize Kafka headers introduced in Kafka version 0.11. CloudEvents currently defines two modes of operation; the related discussion can be followed on [Github](https://github.com/cloudevents/spec/pull/337/files).
```

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
+---
+id: event-serializing
+title: Event Serializing
+sidebar_label: Event Serializing
```
kevinbader:
```suggestion
sidebar_label: Event Serialization
```

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
+---
+id: event-serializing
+title: Event Serializing
```
kevinbader:
```suggestion
title: Event Serialization
```

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
+---
+id: event-serializing
```
kevinbader:
```suggestion
id: event-serialization
```

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
...
+We support [Apache Avro](https://avro.apache.org/) and standard JSON (de)serialization of events for producer as well as for consumer. This is working together with [Cloud Events specification](https://github.com/cloudevents/spec) (versions `0.1` and `0.2`). To make both of these worlds work we are utilizing Kafka headers introduced in version `0.11`. As Cloud Events specify there are 2 modes how to make this combination work. Discussion about this topic can be followed on [Github](https://github.com/cloudevents/spec/pull/337/files).
+
+**Structured content mode**
+
+This mode should be used when JSON format is desired, thus no further serialization and deserialization such as Apache Avro.
+
+> ##### KAFKA EVENT HEADERS
+>
+> ```
+> cloudEvent_contentType: "application/json"
```
kevinbader: shouldn't we use the newer spec here, e.g. `ce-contenttype`?

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
...
+id: event-serializing
+title: Event Serializing
+sidebar_label: Event Serializing
+---
+
+We support [Apache Avro](https://avro.apache.org/) and standard JSON (de)serialization of events for producer as well as for consumer. This is working together with [Cloud Events specification](https://github.com/cloudevents/spec) (versions `0.1` and `0.2`). To make both of these worlds work we are utilizing Kafka headers introduced in version `0.11`. As Cloud Events specify there are 2 modes how to make this combination work. Discussion about this topic can be followed on [Github](https://github.com/cloudevents/spec/pull/337/files).
+
+**Structured content mode**
+
+This mode should be used when JSON format is desired, thus no further serialization and deserialization such as Apache Avro.
```
kevinbader: not sure I get this 😕 

1. if you need json, you can use both modes, the difference is just that in structured mode the context attributes are included in the json too.
1. you can use structured mode with Avro too, right?
1. IIRC we wanted to default to structured mode and JSON encoding - is this also what's implemented? If it is, you should probably mention that here..

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
...
+>    "specversion" : "0.1",
+>    "type" : "com.example.someevent",
+>    "data": {},
+>    ...
+> }
+> ```
+
+**Binary content mode**
+
+This mode should be used when further serialization and deserialization is desired (e.g. Apache Avro).
```
kevinbader: again a bit unclear, probably due to what I hint on with my comment above

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
...
+
+> ##### KAFKA EVENT HEADERS
+>
+> ```
+> cloudEvent_contentType: "avro/binary"
+> cloudEvent_specversion: "0.2"
+> cloudEvent_type: "com.example.someevent"
```
kevinbader: again, we should only use the newer versions here (e.g., ce-contenttype, ce-specversion, ce-type). If we support both that's fine, but in the examples I feel we should be current.

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
...
+>
+> ##### KAFKA EVENT VALUE
+>
+> ```
+> ... application data ...
+> ```
+
+## Apache Avro format
+
+To be compatible across the board Apache Avro is using specific [format](https://docs.confluent.io/current/schema-registry/docs/serializer-formatter.html#wire-format) that needs to be followed.
```
kevinbader: isn't that like the main purpose of a format? :D

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Example
+<0, 0, 0, 0, 1, 5, 3, 8, ...>
+```
+
+### Setup
+
+![event-serializing-avro](assets/event-serializing-avro.png)
+
+Setup for event (de)serialization if fairly straightforward. Besides classic components like web client, RIG and Kafka it needs also `Kafka Schema Registry`. `Kafka Schema Registry` is place where all schemas are stored and retrieved from. As event is consumed from Kafka, RIG fetches schema, caches it in-memory and does deserialization. After that event is sent to web client (or any other client able to connect via WS/SSE). As for producing, RIG fetches schema, caches it in-memory and serialize event. Then it's sent to Kafka.
```
kevinbader:
```suggestion
Adopting Avro for event (de)serialization is fairly straightforward. First you need to run an instance of the `Kafka Schema Registry`, which is a central store for all Avro schemas in use. As an event is consumed from Kafka, RIG fetches the corresponding schema from the registry and deserializes the event with it, caching the schema in the process (in memory). As for producing, RIG again retrieves and caches the schemas used for serializing events.
```

Also, I've wondered, do we support Avro without also using the schema registry? Avro doesn't really require it..

docs/event-serializing.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Producer details
+
+- producer evaluates if serializing is turned on by checking `KAFKA_SERIALIZER` environment variable and if it's value is `avro`
+- If it is, creates headers for Kafka event by appending `cloudEvents_` prefix for every field besides `data` field
+  - **for deep nested values we are using query encoding** since Kafka headers don't support nested values
```
kevinbader: perhaps you should point out that this only applies to so-called CloudEvents "extensions"

integration_tests/kafka_tests/run.sh
mmacai: From what I heard it's better practice to use `printf` instead of `echo` -- https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo

integration_tests/kafka_tests/run.sh
mmacai: Diversity :D

docs/event-serializing.md
mmacai: is there already a newer? crap ...

docs/event-serializing.md
mmacai: I don't follow.

docs/event-serializing.md
mmacai: Not right now, currently we rely on Schema registry, we would have to add support for other type of storing schema such as files on file system.

docs/event-serializing.md
mmacai: Where did you find it?

docs/event-serializing.md
mmacai: If we want to support Avro without schema registry, I would do it iteratively.

docs/event-serializing.md
mmacai: Do we want to try stringified values instead of this monster?

docs/event-serializing.md
mmacai: 1. I'll reword it
2,3. check `Producer details` and `Consumer details` they should provide enough information how it works -- if not, need change

docs/event-serializing.md
mmacai: Will be reworded.

docs/event-serializing.md
mmacai: Can you provide source where they changed it? Checked PR that's referenced in docs and github, but no clear visibility, or I'm blind.

apps/rig_kafka/lib/rig_kafka/serializer.ex
```diff
@@ -48,9 +48,14 @@ defmodule RigKafka.Serializer do
           |> String.replace_prefix(@prefix, "")
           |> String.to_atom()
 
-        decoded_value = Plug.Conn.Query.decode(value)
+        decoded_value =
+          try do
+            Jason.decode!(value)
+          rescue
+            _ -> value
+          end
```
kevinbader: I think it'd be better to use `Jason.decode/1` instead

docs/event-serializing.md
kevinbader: I wasn't thinking about storing the schema anywhere - instead, without a schema registry, the schema would be included in the message itself, for each message, so caching wouldn't even be appropriate.
But I agree, we should do this once the need arises.

docs/event-serializing.md
kevinbader: well Avro is a format specification, so in order to be compatible with Avro - not surprisingly - you gotta follow the format spec. Just thought it's funny, no worries ;)

docs/event-serialization.md
```diff
@@ -0,0 +1,141 @@
...
+- producer sends event with created headers and data (in binary format `<<0, 0, 0, 0, 1, 5, 3, 8, ...>>`) to Kafka
+
+> If `KAFKA_SERIALIZER` is not set to `avro`, producer sets **only** `ce-contenttype` or `ce-contentType` for kafka event
+
+### Consumer details
+
+- when consuming Kafka event, RIG checks headers of such event and removes `ce-` prefix
+- based on headers decides cloud events version and content type
+- In case content type is `avro/binary`, schema ID is taken from event value and deserialized
+- If content type is **not** present, RIG checks for Avro format (`<<0, 0, 0, 0, 1, 5, 3, 8, ...>>`) and attempts for deserialization, otherwise event is sent to client as it is without any deserialization
```
kevinbader: > without any deserialization

it's JSON decoded first - otherwise RIG wouldn't know where to send it to

chadmott:
perhaps I missed something but what happens when a schema in the registry is updated? how will the RIG fetch the latest schema?

mmacai:
@chadmott That's a good point. Right now it would error. Definitely something we have to handle, however I would do it in separate PR with updates to potential issues (I'm pretty sure something will raise up since this is pretty tricky topic).

What is the usual/recommended way to handle this evolution scenario? You have any idea/recommendation?

chadmott:
if your schemes are designed to be fully compatible, which ours are, we have the services poll the registry every five minutes and take latest schema. This works for us, but we are small team and have tight control over the schemas and services. 

That said the _better_ way to do it would be to use a schemaAtVersion/N, and have an endpoint in each service that accepts that call and updates it’s schema reference to a version number.  Update the schema and then execute the REST call to the microservice to use it.

PR #168, opened 2019-02-21
Disable HTTPS on missing param + mount Certfile from one path + don't pack selfsigned certfiles into dockerimg

### LINKED ISSUE #151: Disable HTTPS when cert file is not configured + only use one file path per cert
- We should definitely make sure we load the certs from **one path** -- after all, we're using a single config value but two file locations. So: A user should only mount one cert instead of two.
- A user should be able to mount the cert outside of the `priv` dir (to make it easier to mount a cert into a Docker container; e.g., see the extractor config resolution).
- The docker image shouldn't ship the self-signed cert. It's not even randomly generated, so everybody knows the priv key and I'm a little bit worried people would actually use it in production.

jakobwgnr:
closes #151 

**Attention**:
Not included is following:
* A user should be able to mount the cert outside of the `priv` dir (to make it easier to mount a cert into a Docker container; e.g., see the extractor config resolution).

It seems like the paramteriziation to change phoenix to use a absolute path is not working


jakobwgnr:
Hoi @mmacai 
Could you please have a look on this PR?

apps/rig/lib/rig/config.ex
```diff
@@ -161,4 +186,21 @@ defmodule Rig.Config do
...
+
+  # ---
+
+  defp resolve_path(path) do
+    :code.priv_dir(:rig) |> Path.join(path)
+  end
+
+  # ---
+
+  defp disable_https(_) do
```
mmacai: This is maybe too much, but I see your thinking.

jakobwgnr: @mmacai 

I did this because i think `update_in/3` requires a function to be passed...

```
      # DISABLE HTTPS
      config
      |> update_in([:https], &disable_https/1)
```
I'm not sure if there's a better way to use that function?!

mmacai: Perhaps this should work `fn _ -> false end`, but maybe this one is better, since thanks to function name you know what's going on, so we can keep it (doesn't matter that much at the end of the day).

jakobwgnr: aah - true that :) OK. I'll keep like that

jakobwgnr:
Hi @mmacai 
Hmm - It seems like `update_docs.ex` does use "nil" as default... 
`cur_default_str = Map.get(defaults, key, nil) |> inspect()`

Therfore it's updating the empty string to nil...

If I change to `cur_default_str = Map.get(defaults, key, "") |> inspect()`
then the default for HTTPS_CERTFILE, etc. is fine and doesn't change... BUT Defaults from other ENVs change --> E.g. KAFKA_RESTART_DELAY_MS

Unfortunately I can't use nil, as phoenix isn't capable to work with nil as default... I have to use "" ...

As this isn't really related to that issue we probably should create a separate issue to handle this case.
`update_docs.ex` should be capable to handle "" AND nil as a default?!

What do you think? @kevinbader

jakobwgnr:
@mmacai BTW - I've updated and committed the docu files

mmacai:
Agree, let's create an issue for that.

docs/tutorial.md
```diff
@@ -11,13 +11,24 @@ In this tutorial we use [HTTPie](https://httpie.org/) for HTTP requests, but of
 To get started, run our Docker image using this command:
 
 ```bash
-$ docker run -p 4000:4000 -p 4010:4010 -e HTTPS_CERTFILE=cert/selfsigned.pem -e HTTPS_KEYFILE=cert/selfsigned_key.pem accenture/reactive-interaction-gateway:2.0.2
+$ docker run \
+-e HTTPS_CERTFILE=cert/selfsigned.pem \
```
mmacai: I would omit also these 2 guys since we have note about SSL below.

mmacai:
Please also resolve conflict in `apps/rig_api/config/config.exs`

docs/tutorial.md
jakobwgnr: agree. Done

jakobwgnr:
Hi @mmacai 
Updated tutorial + resolved merge conflict.

kevinbader:
> Hmm - It seems like `update_docs.ex` does use "nil" as default...
> `cur_default_str = Map.get(defaults, key, nil) |> inspect()`
> 
> Therfore it's updating the empty string to nil...
> 
> If I change to `cur_default_str = Map.get(defaults, key, "") |> inspect()`
> then the default for HTTPS_CERTFILE, etc. is fine and doesn't change... BUT Defaults from other ENVs change --> E.g. KAFKA_RESTART_DELAY_MS
> 
> Unfortunately I can't use nil, as phoenix isn't capable to work with nil as default... I have to use "" ...
> 
> As this isn't really related to that issue we probably should create a separate issue to handle this case.
> `update_docs.ex` should be capable to handle "" AND nil as a default?! 

@jakobwgnr From the script's perspective, having nil as a default seems to make sense. After all, if you need `""` as the default value you can always add it to the env var definition: `{:system, "MY_VAR", ""}`. Am I missing the point here?

jakobwgnr:
> > Hmm - It seems like `update_docs.ex` does use "nil" as default...
> > `cur_default_str = Map.get(defaults, key, nil) |> inspect()`
> > Therfore it's updating the empty string to nil...
> > If I change to `cur_default_str = Map.get(defaults, key, "") |> inspect()`
> > then the default for HTTPS_CERTFILE, etc. is fine and doesn't change... BUT Defaults from other ENVs change --> E.g. KAFKA_RESTART_DELAY_MS
> > Unfortunately I can't use nil, as phoenix isn't capable to work with nil as default... I have to use "" ...
> > As this isn't really related to that issue we probably should create a separate issue to handle this case.
> > `update_docs.ex` should be capable to handle "" AND nil as a default?!
> 
> @jakobwgnr From the script's perspective, having nil as a default seems to make sense. After all, if you need `""` as the default value you can always add it to the env var definition: `{:system, "MY_VAR", ""}`. Am I missing the point here?

@kevinbader yep - I did. I have defined the default in the env var definition & it is working fine. It's just that the script is automatically updating the ops-guide to nil as it seems it doesn't identify `""` 
so we could just keep "nil" as the default value in the ops-guide as it anyways doesn't really matter

PR #174, opened 2019-03-01
Fixed proxy parsing when atom was expected

mmacai:
In master this would fail: `PROXY_CONFIG_FILE=proxy/proxy.test.json mix phx.server`


mmacai:
Gone.

PR #175, opened 2019-03-01
#173 minor updates to doku

### LINKED ISSUE #173: Tutorial Improvements
I was just going through the documentation and I think we could slightly improve the tutorial. 

Following things I could think of:
- Describing the use case we are showing
- Describing the artifacts we are using/mocking in this tutorial (RIG, the frontend, the backend)
- Clearly mention for each step which artifact we're working on right now
- Remove step 6 from the workflow as it actually doesn't fit there. It should be mentioned that this is how you would implement above's steps in a JS frontend?

What do you guys think? @kevinbader @mmacai

jakobwgnr:
closes #173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Proxy] Parsing of JSON files in proxy module. Expected that `api.id` is an atom, but when using files it's a string.
+- [Docs] Fixed channels example with latest RIG API changes.
 
 <!-- ### Security -->
 
--- a/examples/channels-example/app.docker-compose.yml
+++ b/examples/channels-example/app.docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: "3"
 services:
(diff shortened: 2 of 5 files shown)


jakobwgnr:
Hi @kevinbader 
Seems like there were committed some files automatically during rebase...

Can you remove them? as they are actually not part of this PR

apps/rig/lib/mix/tasks/update_docs.ex
```diff
@@ -14,18 +14,11 @@ defmodule Mix.Tasks.UpdateDocs do
...
-    # (as only then we'll find the correct default values):
-    if Project.umbrella?() and Mix.env() == :prod do
-      if File.exists?(@target_path) do
-        update_file(@target_path)
-      else
-        Logger.info(fn -> "Not updating documentation (file not present: #{@target_path})" end)
-      end
+    # Only run when compiling the whole umbrella application in DEV:
+    if Project.umbrella?() and Mix.env() == :dev do
+      update_file("./docs/rig-ops-guide.md")
```
kevinbader: hm can you please re-do the merge? seems you're removing recent changes here..

PR #184, opened 2019-03-08
https docu + updated cert file mounting

### LINKED ISSUE #182: Absolute path for https cert file + https doku
Similar to proxy_config_file the https cert file should be mountable with an absolute path.

Additionally, https should be documented using an separate page in the guides

jakobwgnr:
Closes #182 

@kevinbader 
Could you have a look on this one? When I now run rig with docker + mounting the files as proposed in the new https.md I don't get any errors + as of logs also cowboy with correct port is started.

BUT: When i try to call rig with https I get a connection refused.

when I run rig using mix phx.server (using standard dev parameters) I'm able to call rig on https.

I'm not sure if I'm doing something wrong on mounting the files or missing some param on docker run? Or is it sth else?

### DIFF
--- a/apps/rig/lib/rig/config.ex
+++ b/apps/rig/lib/rig/config.ex
@@ -94,7 +94,7 @@ defmodule Rig.Config do
 
   @spec parse_json_env(String.t()) :: {:ok, any} | {:error, :syntax_error, any}
   def parse_json_env(path_or_encoded) do
-    with {:error, reason1} <- from_file(path_or_encoded),
+    with {:error, reason1} <- decode_json_file(path_or_encoded),
          {:error, reason2} <- from_encoded(path_or_encoded) do
       {:error, :syntax_error, [reason1, reason2]}
     else
@@ -117,8 +117,8 @@ defmodule Rig.Config do
     else
       # UPDATE https_config to add priv/ folder to path
       config
-      |> update_in([:https, :certfile], &resolve_path/1)
-      |> update_in([:https, :keyfile], &resolve_path/1)
+      |> update_in([:https, :certfile], &update_https_path/1)
+      |> update_in([:https, :keyfile], &update_https_path/1)
       |> update_in([:https, :password], &String.to_charlist/1)
     end
   end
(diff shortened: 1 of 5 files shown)


apps/rig/lib/rig/config.ex
```diff
@@ -189,8 +189,18 @@ defmodule Rig.Config do
 
   # ---
 
+  @spec resolve_path(String.t()) :: Struing.t()
```
kevinbader: There's a typo. But I'd like to suggest a slightly different signature anyway - what about:

```suggestion
  @spec resolve_path(path) :: {:ok, path} | {:error, any} when path: String.t()
```

- Then, use `resolve_path/1` also in `from_file`, and rename `from_file` to `decode_json_file`.
- Instead of logging the error here, log at the call site, that is, in `check_and_update_https_config/1`. This makes the function more re-usable, since `resolve_path/1` cannot decide whether or not logging make sense (also, the fact that the log message contains "HTTPS" basically prevents any possible re-use of the function). **In any case, `check_and_update_https_config/1` should handle a non-resolvable file path (I'd suggest aborting RIG with the error message).** Note that currently, the function returns `:ok` in case the file path cannot be resolved, which is quite misleading.

docs/https.md
```diff
@@ -0,0 +1,35 @@
...
+$ export HTTPS_CERTFILE=cert/selfsigned.pem
+$ export HTTPS_KEYFILE=cert/selfsigned_key.pem
+$ mix phx.server
+```
+
+WARNING: only use the generated certificate for testing in a closed network environment, such as running a development RIG instance on localhost. For production, staging, or testing on the public internet, obtain a proper certificate, for example from [Let’s Encrypt](https://letsencrypt.org/).
+
+For production you should use proper HTTPS certificates instead (for that reason the Docker image comes without certificates).
+
+1. Create your own not-selfsigned certificate
```
kevinbader:
```suggestion
1. Create your own certificate
```

Depending on the situation, a self-signed certificate might actually _okay_. The important thing here is that no-one uses the one from the repo :)

apps/rig/lib/rig/config.ex
jakobwgnr: updated - please check

PR #185, opened 2019-03-18
169 - fixed broken examples using jwt inferred subscriptions

### LINKED ISSUE #169: Broken examples
## Summary and context

As identified during PR #166 the two following examples are not working right now.
´examples/sse-demo-jwt-extractors.html´
´examples/sse-demo-jwt-extractors-conn.html´

## How to reproduce

Try out the examples

 - Host OS: Windows 
 - Frontend chrome, safari
 - RIG version as shown on startup 2.1.0

mmacai:
Closes #169 

Examples were using `Bearer ...` format in query headers which was wrong, now it's fixed and inferred subscriptions are covered with tests.

- all examples should work now
- all types of tests - unit, smoke, integration should pass


apps/rig/test/rig/subscriptions_test.exs
```diff
@@ -22,18 +22,18 @@ defmodule RigInboundGateway.ImplicitSubscriptions.JwtTest do
         "fullname" => %{
           "event" => %{"json_pointer" => "/data/fullname"},
           "jwt" => %{"json_pointer" => "/fullname"},
-          "stable_field_index" => 1
+          "stable_field_index" => 2
```
kevinbader: the `stable_field_index` is per event type, so this shouldn't be necessary

PR #190, opened 2019-03-30
Swagger & Kafka group ID fixes

mmacai:
- fixed missing id field for message swagger spec
- fixed random generation of Kafka group id

Test:
- `mix test` & integration tests should pass
- run rig locally - sending `/message` form `http://localhost:4010/swagger-ui` should work on ootb
- same should work from docker container

- run rig with Kafka => check consumer => all should have consumer group set to `rig` - can be changed via env var

### DIFF
--- a/apps/rig_api/lib/rig_api/controllers/message_controller.ex
+++ b/apps/rig_api/lib/rig_api/controllers/message_controller.ex
@@ -60,6 +60,14 @@ defmodule RigApi.MessageController do
           description("The broadcasted CloudEvent according to the CloudEvents spec.")
 
           properties do
+            id(
+              :string,
+              "ID of the event. The semantics of this string are explicitly undefined to ease \
+              the implementation of producers. Enables deduplication.",
+              required: true,
+              example: "A database commit ID"
+            )
+
             specversion(
               :string,
               "The version of the CloudEvents specification which the event uses. This \
--- a/apps/rig_kafka/lib/rig_kafka/config.ex
+++ b/apps/rig_kafka/lib/rig_kafka/config.ex
@@ -51,7 +51,7 @@ defmodule RigKafka.Config do
       consumer_topics: Map.get(config, :consumer_topics, []),
       server_id: Map.get(config, :server_id) || String.to_atom("rig_kafka_#{uuid}"),
       client_id: Map.get(config, :client_id) || String.to_atom("brod_client_#{uuid}"),
-      group_id: Map.get(config, :group_id) || "brod_group_rig",
+      group_id: Map.get(config, :group_id),
       ssl: Map.get(config, :ssl),
       sasl: Map.get(config, :sasl)
     }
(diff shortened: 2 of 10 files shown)


apps/rig_api/lib/rig_api/controllers/message_controller.ex
```diff
@@ -60,6 +60,14 @@ defmodule RigApi.MessageController do
...
 
           properties do
+            id(
+              :string,
+              "ID of the event. The semantics of this string are explicitly undefined to ease \
+              the implementation of producers. Enables deduplication.",
+              required: true,
+              example: "A database commit ID"
+            )
+
```
kevinbader: isn't `source` required too?

apps/rig_kafka/lib/rig_kafka/config.ex
```diff
@@ -51,7 +51,7 @@ defmodule RigKafka.Config do
       consumer_topics: Map.get(config, :consumer_topics, []),
       server_id: Map.get(config, :server_id) || String.to_atom("rig_kafka_#{uuid}"),
       client_id: Map.get(config, :client_id) || String.to_atom("brod_client_#{uuid}"),
-      group_id: Map.get(config, :group_id) || "brod_group_#{uuid}",
+      group_id: Map.get(config, :group_id) || "brod_group_rig",
```
kevinbader: Why is the default here different from the default in `config.exs`? Actually, I'm not sure a default makes sense here in the first place - after all, we already define defaults in the config files.

website/i18n/en.json
```diff
@@ -37,10 +37,6 @@
         "title": "Metrics Details",
         "sidebar_label": "Metrics Details"
       },
-      "https": {
-        "title": "HTTPS",
-        "sidebar_label": "HTTPS"
-      },
```
kevinbader: not sure why this is here

integration_tests/kafka_tests/run.sh
```diff
@@ -22,7 +22,7 @@ while ! is_kafka_ready; do
     sleep 1
 done
 
-check whether Kafka is now ready or the loop has been aborted:
+# check whether Kafka is now ready or the loop has been aborted:
```
kevinbader: we don't run them often, do we? :D

apps/rig_api/lib/rig_api/controllers/message_controller.ex
mmacai: Should be, but apparently not in our code. At least I didn't see any error when using `/messages` endpoint.

apps/rig_kafka/lib/rig_kafka/config.ex
mmacai: Left it there due to tests, I'll remove it and put in test config.

integration_tests/kafka_tests/run.sh
mmacai: Actually I made some improvements in Avro PR -- so should be better in near future :D.

apps/rig_api/lib/rig_api/controllers/message_controller.ex
kevinbader: created an issue for this: #191

PR #192, opened 2019-04-05
More documentation

mmacai:
~~This is WIP - should cover most of the functionality.~~

- added more docs for API gateway
- event streams
- scaling
- revamped kubernetes configs
- fixed swagger-ui in production docker image


mmacai:
@kevinbader you can take a look -- still some changes coming, but should be fine to do first round of feedback.

docs/event-streams.md
```diff
@@ -0,0 +1,186 @@
+---
+id: event-streams
+title: Event Streams
+sidebar_label: Event Streams
+---
+
+Real time part of RIG is driven by events (Kafka/Kinesis) -- to make it work there is some configuration required.
+
+Where is RIG using event streams:
```
kevinbader:
```suggestion
Where RIG uses event streams:
```

docs/event-streams.md
```diff
@@ -0,0 +1,186 @@
...
+-e KAFKA_BROKERS=kafka:9092 \
+-e KAFKA_SOURCE_TOPICS=my-topic \
+-e PROXY_KAFKA_RESPONSE_TOPICS=my-proxy-topic \
+-e KAFKA_GROUP_ID=my-group-id \
+accenture/reactive-interaction-gateway
+```
+
+### Change producer topics
+
+Rig can produce "monitoring" events when HTTP endpoint is called in proxy and as a target of HTTP endpoint itself, see [API Gateway docs](./api-gateway#sync).
```
kevinbader:
```suggestion
RIG can produce "monitoring" events on incoming HTTP requests, see [API Gateway docs](./api-gateway#sync).
```

docs/event-streams.md
```diff
@@ -0,0 +1,186 @@
...
+
+```bash
+# Kinesis disabled
+docker run accenture/reactive-interaction-gateway
+
+# Kinesis enabled
+docker run -e KINESIS_ENABLED=1 accenture/reactive-interaction-gateway
+
+# Configure AWS region
+docker run -e KINESIS_ENABLED=1 -e KINESIS_AWS_REGION=eu-west-3 accenture/reactive-interaction-gateway
```
kevinbader: wouldn't you need to use the aws docker-tag?

docs/event-streams.md
```diff
@@ -0,0 +1,186 @@
...
+docker run \
+-e KINESIS_ENABLED=1 \
+-e KINESIS_STREAM=my-stream \
+-e KINESIS_APP_NAME=my-app_name \
+accenture/reactive-interaction-gateway
+```
+
+### Change producer streams
+
+Rig can produce events as a target of HTTP endpoint call, see [API Gateway docs](./api-gateway#sync).
```
kevinbader:
```suggestion
RIG can produce events as a target of HTTP endpoint call, see [API Gateway docs](./api-gateway#sync).
```

docs/api-gateway.md
```diff
@@ -91,3 +92,230 @@ Note that this way we don't need a Docker volume, which might work better in you
...
+        "target": "kafka",
+        "response_from": "kafka"
+      }]
+    }
+  },
+  ...
+}]
+```
+
+> Important fields are `target` and `response_from`. Both of them describe what event stream to use. Currently it's either Kafka or Kinesis. It's recommended to use `POST` http method. `path` name is up to you.
```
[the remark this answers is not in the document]
mmacai: Just for convenience I guess, since method like GET doesn't make much of a sense, but ye still can be PUT or PATCH.

docs/api-gateway.md
```diff
@@ -91,3 +92,230 @@ Note that this way we don't need a Docker volume, which might work better in you
...
+
+Dynamic values in `path` have to be wrapped in curly braces. Value inside curly braces is up to you.
+
+## Publishing to event streams
+
+As an alternative to standard HTTP to HTTP communication, we offer a way how to produce event to Kafka/Kinesis via HTTP request. This means, that frontend can still send regular HTTP request and RIG will publish content of this request to Kafka topic or Kinesis stream. We distinguish between two types of such request -- sync and async.
+
+### Sync
+
+Sync means that as you send HTTP request to RIG, it publishes event to Kafka/Kinesis and waits for the acknowledgment from consumer. Consumer has to receive event with the same correlation ID. If such event is consumed, HTTP process is notified and client gets back response, otherwise it's timeout. The correlation ID is attached to Cloud events extension in published event called `rig`. In your backend systems you have to make sure that this extension field will be included also in the event that should be consumed by RIG and finish the process.
```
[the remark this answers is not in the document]
mmacai: Ye, I'm not a fan of sync/async as well, I'll revisit it.

docs/event-streams.md
mmacai: True, good point!

docs/api-gateway.md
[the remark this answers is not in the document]
kevinbader: Let's say you have some services that compute something together when one of them sees a request. At the end, the result is written to a Kafka topic. The process is asynchronous, but the type of request might be an idempotent retrival, for which GET would be the best fit in terms of HTTP verbs.

docs/api-gateway.md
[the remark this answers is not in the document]
mmacai: Rephrased it a bit, let me know what you think. Tried to avoid sync and async words as much as possible.

docs/api-gateway.md
```diff
@@ -91,3 +92,228 @@ Note that this way we don't need a Docker volume, which might work better in you
...
+}
+```
+
+As you can see, it needs 2 fields -- `event` and `partition`. `event` holds event itself (have to comply with Cloud events spec) and `partition` describes partition key to use.
+
+Topic/stream configuration is handled by environment variables. See `PROXY_KAFKA_*` and `PROXY_KINESIS_*` variables in [Operator's Guide](./rig-ops-guide.md).
+
+### Wait for response
+
+Publishing to event stream is asynchronous action (it's not request/response mechanism like HTTP is). Endpoint above can be configured, so that RIG waits for different event with the same correlation ID as the published event had. If such event is consumed, HTTP process is notified and client gets back response, otherwise it's timeout. The correlation ID is attached to Cloud events extension in published event called `rig`. In your backend systems you have to make sure that this extension field will be included also in the event that should be consumed by RIG and finish the process.
```
kevinbader:
```suggestion
Sometimes it makes sense to provide a simple request-response API to something that runs asychronously on the backend. For example, let's say there's a ticket reservation process that takes 10 seconds in total and involves three different services that communicate via message passing. For an external client, it may be simpler to wait 10 seconds for the response instead of polling for a response every other second.

A behavior like this can be configured using an endpoints' `response_from` property. When set to `kafka`, the response to the request is not taken from the `target` (e.g., for `target` = `http` this means the backend's HTTP response is ignored), but instead it's read from a Kafka topic. In order to enable RIG to correlate the response from the topic with the original request, RIG adds a correlation ID to the request (using a query parameter in case of `target` = `http`, or backed into the produced CloudEvent otherwise). Back-end services that work with the request need to include that correlation ID in their response; otherwise, RIG won't be able to forward it to the client (and times out).
```

PR #209, opened 2019-04-28
Implement HTTP Transport Binding for CloudEvents v0.2

### LINKED ISSUE #153: CloudEvents HTTP transport binding
RIG does not yet implement the [HTTP transport binding spec](https://github.com/cloudevents/spec/blob/master/http-transport-binding.md).

> The receiver of the event can distinguish between the two modes by inspecting the Content-Type header value. If the value is prefixed with the CloudEvents media type application/cloudevents, indicating the use of a known event format, the receiver uses structured mode, otherwise it defaults to binary mode.

### Structured mode (content type = application/json)

[To be implemented](https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#31-binary-content-mode). Currently, we do binary mode here.

### Binary mode for JSON (application/cloudevents+json)

That's what we do right now, except for the header, which we expect to be "application/json".

kevinbader:
## Description

Implements the [HTTP Transport Binding for CloudEvents spec in version 0.2](https://github.com/cloudevents/spec/blob/v0.2/http-transport-binding.md). A special fallback to "structured mode" in case the content type is "application/json" and the "ce-specversion" header is not set ensures this change is backward compatible with existing setups.

Closes #153.

- [ ]  ~~Apply the changes to what's introduced by #215~~ <-- postponed until the need arises
- [x]  Make sure the examples work.
- [x]  Update the docs.

## What to look out for

- I think it's important for you to know about the change. @mmacai **please verify the code changes but also the documentation updates match what you've implemented for Avro.**
- I'm aware that the event controller and the message controller are almost identical (expect for the swagger documentation). I'm reluctant to factor the implementations out into a common module, because the files are currently part of different umbrella apps. Suggestions?
- The Mox mock stopped working, apparently. The related assertions are not that important, but still, it's annoying. Any insights welcome.
- Check that the code works on your machine. I've added a couple of tests and it _seems_ to work, but I had some media-type related fights with Phoenix/Plug, so I'm not too confident about it, unfortunately.
- Suggest implementation-code design/structure/readability improvements.
- Suggest test-code design/structure/readability improvements.

## Process

Have fun!


docs/minimal-sync-to-async-proxy-config.json
```diff
@@ -0,0 +1,27 @@
+[
```
mmacai: Is this used somewhere? If so, looks outdated.

docs/user-authorization.md
```diff
@@ -0,0 +1,31 @@
...
+---
+
+RIG supports the [JSON Web Tokens](https://jwt.io/) (JWT) standard for authorizing users when forwarding HTTP requests, or when handling event subscription or submission requests.
+
+Note that RIG does _not_ do _authentication_ - you will need to provide a service for that yourself. In terms of tokens, this means that RIG won't ever issue JWTs itself; instead, RIG only checks their validity using a shared secret key or a public key, depending on the algorithm used. Please refer to the [Operator's Guide](rig-ops-guide) for the corresponding configuration options.
+
+Also note the following:
+
+- For incoming HTTP requests, JWT validation can be enabled on a per endpoint basis using the `secured` option.
+- Event subscriptions can be secured using JWT validation by setting `SUBSCRIPTION_CHECK` to `jwt_validation`.
```
mmacai: I would mention here and below that those are env vars or link to ops guide.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/event_controller.ex
```diff
@@ -21,18 +26,155 @@ defmodule RigInboundGatewayWeb.V1.EventController do
...
+          """)
+      end
+    else
+      ce_specversion_headers ->
+        conn
+        |> put_status(:bad_request)
+        |> text("""
+        Your request looks like it's using the CloudEvents binary content mode, \
+        but the ce-specversion header is #{inspect(ce_specversion_headers)} \
+        instead of [\"0.2\"].\
```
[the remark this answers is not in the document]
kevinbader: 0.1 is still supported for structured mode but afaik there never was support for 0.1 in binary mode

docs/user-authorization.md
kevinbader: I've put another link to the ops guide next to it.

mmacai:
My todo list to test:

- [ ] html examples
- [ ] channels example
- [ ] api gateway docs examples
- [ ] avro

@kevinbader should be enough to make sure it's alright.

docs/api-gateway.md
```diff
@@ -140,9 +140,9 @@ Configuration of such API endpoint might look like this:
...
 ```
 
-> Important field is `target`. `target` describes what event stream to use. Currently it's either Kafka or Kinesis. HTTP method and `path` name are up to your choice.
+Note that the `target` field is set to `kafka` (for Kinesis use `kinesis`).
 
-This type of endpoint requires specific HTTP request body:
+The endpoint expects the following request format:
 
 ```json
 {
```
mmacai: Perhaps here, we should mention current format/approach. This one is deprecated if I understood correctly.

kevinbader: good catch, will update that

PR #214, opened 2019-05-15
170 - re-enabled request logger and made compatible with Avro

### LINKED ISSUE #170: Broken kafka request logger
## Summary and context

We have wrong configuration for kafka producer in `lib/rig_inbound_gateway/request_logger/kafka.ex`

## How to reproduce

`mix phx.server`

Warning:

```
==> rig_inbound_gateway
Compiling 15 files (.ex)
warning: function Rig.Kafka.produce/3 is undefined (module Rig.Kafka is not available)
  lib/rig_inbound_gateway/request_logger/kafka.ex:55
```

 - Host OS: Mac OSX
 - RIG version: 2.1.0

### LINKED ISSUE #201: Integration tests hardening
## Summary and context

Some of the integration tests for Kafka fail from time to time, because consumer process creation takes a bit longer and thus test times out.

## How to reproduce

```
cd ../integration_tests/kafka_tests/ && ./run.sh
```

 - Docker
 - Reactive Interaction Gateway 2.1.1

mmacai:
## Description

Fixed API Gateway's request logger to work with latest Kafka client/interface & added integration with Avro.

Split integration tests into 2 parts - Kafka without Avro & Kafka with Avro -- reason is easier configuration and less issues in tests themselves.

Closes #170 
Closes #201 

What's missing:

- [x] working integration test for request logger with Avro

## What to look out for

- Check docs and try to run RIG with the instructions found there. (might be that it's not informative enough)
- There is an issue with request logger integration test and Avro (the commented one) - it's due to the fact that as RIG starts it spawns producer process for request logger and if I try to change env var ad-hoc in test it can't be overridden (since process already started). - @kevinbader any thoughts?
  - Solved by splitting integration tests into 2 parts - with and without avro -- this should also increase integration tests stability

## Process

Have fun!


mmacai:
Ready for review.

apps/rig_tests/test/avro/avro_test.exs
```diff
@@ -59,16 +57,15 @@ defmodule RigTests.Avro.AvroTest do
       })
 
     kafka_config = kafka_config()
+    :timer.sleep(5_000)
```
kevinbader: Why is this (and the other `sleep`s below) needed?

apps/rig_tests/test/proxy/publish_to_event_stream/kafka_test.exs
```diff
@@ -160,6 +160,8 @@ defmodule RigTests.Proxy.PublishToEventStream.KafkaTest do
         "partition" => "test_key"
       })
 
+    :timer.sleep(5_000)
```
kevinbader: does setting up the route take so long? please add a comment to describe why we're waiting here or remove the sleep altogether if possible :)

apps/rig_tests/test/proxy/request_logger/kafka_test.exs
```diff
@@ -0,0 +1,185 @@
...
+        }
+      })
+
+    headers = [{"content-type", "application/json"}]
+    HTTPoison.post!(rig_api_url, body, headers)
+
+    # The client calls the proxy endpoint:
+    request_url = rig_proxy_url <> endpoint_path
+
+    :timer.sleep(5_000)
```
kevinbader: see other comment

apps/rig_tests/test/proxy/request_logger/kafka_test.exs
```diff
@@ -0,0 +1,185 @@
...
+
+    %HTTPoison.Response{status_code: res_status, body: res_body} =
+      HTTPoison.post!(request_url, "", headers)
+
+    assert res_status == 200
+    assert res_body == "{\"status\":\"ok\"}"
```
kevinbader: why are you asserting status code and body here? seems the test should only show that the request has been logged - there are other tests that make sure the request actually went through (imho putting a focus on what the test actually tests makes the test a lot clearer)

apps/rig_tests/test/proxy/request_logger/kafka_test.exs
```diff
@@ -0,0 +1,185 @@
...
+
+    assert_receive received_msg, 10_000
+    received_msg_map = Jason.decode!(received_msg)
+
+    System.delete_env("REQUEST_LOG")
+
+    # Endpoint
+    assert get_in(received_msg_map, ["data", "endpoint", "id"]) == endpoint_id
```
kevinbader: I wonder if it'd make sense to have a type field of some sort that tells us whether this is actually a log entry or just some json that happens to have the same fields. What do you think?

docs/api-gateway.md
```diff
@@ -18,31 +18,35 @@ To pass the configuration at startup, RIG uses an environment variable called `P
...
+]
 ```
 
 This defines a single service called "my-service". The URL is read from an given environment variable in this case (`use_env: true`). Because we want to run RIG inside a Docker container, we cannot use `localhost`. Instead, we can use `host.docker.internal` within the container to refer to the Docker host. This way, the service URL is resolved to `http://host.docker.internal:3000`. The service has one endpoint called "my-endpoint" at path `/`, which forwards `GET` requests to the same path (`http://host.docker.internal:3000/`).
 
 As a demo service, we use a small Node.js script:
 
 ```js
-const http = require("http");
+const http = require('http');
```
kevinbader: accident?

apps/rig_tests/test/avro/avro_test.exs
mmacai: From my observations, from time to time clients need some time to be in ready state and since we miss proper ready function for that, I chose this -- didn't find better and stable approach.

apps/rig_tests/test/proxy/request_logger/kafka_test.exs
mmacai: Agree, probably a bad habit.

apps/rig_tests/test/proxy/request_logger/kafka_test.exs
mmacai: You mean cloud events' type?

docs/api-gateway.md
mmacai: Ye, VS code autoformatting ^ ^.

apps/rig_tests/test/proxy/request_logger/kafka_test.exs
kevinbader: of course, didn't really consider that this is _already_ a cloudevent ^^ yes, the cloudevent's type seems appropriate

apps/rig_tests/test/avro/avro_test.exs
kevinbader: aah _that_ issue...

mmacai:
> Changes look good for the most part! What could be improved:
> 
> * If RIG is started with `REQUEST_LOG=asdf`, I'd expect RIG to fail on startup. Instead, RIG returns `500 Internal Server Error` when hit with a request, without emitting a warning of any kind.
> * RIG fails the same way for `REQUEST_LOG=`, which really should only disable request logging.
> * RIG also fails for `REQUEST_LOG=kafka` in case Kafka is not configured.
>   
>   1. RIG should fail on startup if this is set and Kafka is not enabled.
>   2. If Kafka _is_ enabled but there's a connection issue, the request logger should emit a warning but the service response should _not_ be affected - the client shouldn't see a 500 just because its request couldn't be logged successfully.

Hmm, do you think it's necessary to shutdown entire RIG when env var is misconfigured or kafka is not set, but logger (with kafka value) is?

In the first case, I guess it makes sense - Kafka works in a similar way when you misconfigure something.

In the second case, I'm not sure -- warning seems enough for me.

kevinbader:
> Hmm, do you think it's necessary to shutdown entire RIG when env var is misconfigured or kafka is not set, but logger (with kafka value) is?

I wouldn't shut down a running instance, but I would prevent starting a new one with this configuration.

mmacai:
> > Hmm, do you think it's necessary to shutdown entire RIG when env var is misconfigured or kafka is not set, but logger (with kafka value) is?
> 
> I wouldn't shut down a running instance, but I would prevent starting a new one with this configuration.

Added my first shot, but you can more or less see the direction in git diff.

In general, goal is:

- if `REQUEST_LOG` is set to something random => exit
- if `REQUEST_LOG=` => exit
- if `REQUEST_LOG=kafka`, but `KAFKA_BROKERS` not set => exit
- if `REQUEST_LOG=kafka`, `KAFKA_BROKERS` set, but there are Kafka connection issues => HTTP response should be ok and produce fail
  - changed logging from `call` to `cast` as in my opinion shouldn't block the HTTP response to client

kevinbader:
Please rename `RIG.ConfigValidation` to something related to the logger. I feel it's better to group validations with the feature they're related to, rather than creating a central module that potentially validates _everything_.

> * if `REQUEST_LOG` is set to something random => exit

👍 

> * if `REQUEST_LOG=` => exit

Above I wrote "RIG fails the same way for REQUEST_LOG=, which really should only disable request logging." - still feel the same way. Why do you prefer exit in this case?

> * if `REQUEST_LOG=kafka`, but `KAFKA_BROKERS` not set => exit

👍 

> * if `REQUEST_LOG=kafka`, `KAFKA_BROKERS` set, but there are Kafka connection issues => HTTP response should be ok and produce fail

👍 if with "produce fail" you mean "produce warning".

>   * changed logging from `call` to `cast`, as in my opinion it shouldn't block the HTTP response

👍

mmacai:
Added filtering for env var value that should prevent failing on cases like `REQUEST_LOG=` or `REQUEST_LOG=kafka,`.

PR #215, opened 2019-05-16
213 - Allow to send async response using internal HTTP endpoint

### LINKED ISSUE #213: Allow to send async response using HTTP request against internal port
The goal is to create a new response_from target that basically says “don’t use the sync http response but also don’t look for the response from Kafka; instead, wait for the response to be submitted through an HTTP call against the internal port”.

mmacai:
## Description

Added new internal `POST` endpoint `/v1/responses` to ack pending reverse proxy request. Added new `response_from` option `http_internal` to await ack from mentioned endpoint.

Closes #213 

## What to look out for

- Check docs and swagger
- Check naming - not sure if `http_internal` is the best choice, but didn't think about anything better, short and clear enough for everyone
- Check implementation and integration test (decided to put it alongside existing `http` scenario)
- @chadmott verify if following solution fulfils your needs

You can also test it via:

```bash
# terminal 1
config="[{\"id\":\"my-service\",\"version_data\":{\"default\":{\"endpoints\":[{\"id\":\"my-endpoint\",\"method\":\"GET\",\"path\":\"\/http\",\"response_from\":\"http_internal\"}]}},\"proxy\":{\"use_env\":true,\"target_url\":\"API_HOST\",\"port\":3000}}]"
# terminal 1
PROXY_CONFIG_FILE="$config" PROXY_HTTP_RESPONSE_TIMEOUT=30000 mix phx.server
# terminal 2
node examples/api-gateway/service.js
# terminal 3
curl "http://localhost:4000/http"
# terminal 4 - copy&paste correlation ID from terminal 2
curl -X POST "http://localhost:4010/v1/responses" -H "accept: application/json" -H "content-type: application/json" -d "{ \"rig\": { \"correlation\": \"g2dkAA1ub25vZGVAbm9ob3N0AAADpgAAAAAA\" }}"
```

## Process

Have fun!

### DIFF
--- a/apps/rig_api/lib/rig_api/controllers/responses_controller.ex
+++ b/apps/rig_api/lib/rig_api/controllers/responses_controller.ex
@@ -1,0 +1,78 @@
+defmodule RigApi.ResponsesController do
+  require Logger
+
+  use RigApi, :controller
+  use PhoenixSwagger
+
+  alias Rig.Connection.Codec
+
+  action_fallback(RigApi.FallbackController)
+
+  swagger_path :create do
+    post("/v1/responses")
+    summary("Submit a message, to be sent to correlated reverse proxy request.")
+    description("Allows you to submit a message to RIG using a simple, \
+    synchronous call. Message will be sent to correlated reverse proxy request.")
+
+    parameters do
+      messageBody(
+        :body,
+        Schema.ref(:Response),
+        "Response",
+        required: true
+      )
+    end
+
+    response(202, "Accepted - message sent to correlated reverse proxy request")
+    response(400, "Bad Request: Failed to parse request body :parse-error")
+  end
+
+  @doc """
+  Accepts message to be sent to correlated HTTP process.
+
+  Note that body has to contain following field `"rig": { "correlation": "_id_" }`.
+  """
+  def create(conn, message) do
+    with {:ok, rig_metadata} <- Map.fetch(message, "rig"),
+         {:ok, correlation_id} <- Map.fetch(rig_metadata, "correlation"),
+         {:ok, deserialized_pid} <- Codec.deserialize(correlation_id),
+         {:ok, encoded_body} <- Jason.encode(message) do
+      Logger.debug(fn ->
+        "HTTP response via internal HTTP to #{inspect(deserialized_pid)}: #{inspect(message)}"
+      end)
+
+      send(deserialized_pid, {:response_received, encoded_body})
+      send_resp(conn, :accepted, "message sent to correlated reverse proxy request")
+    else
+      err ->
+        Logger.warn(fn -> "Parse error #{inspect(err)} for #{inspect(message)}" end)
+
+        conn
+        |> put_status(:bad_request)
+        |> text("Failed to parse request body: #{inspect(err)}")
+    end
+  end
+
+  def swagger_definitions do
+    %{
+      Response:
+        swagger_schema do
+          title("Response object")
+          description("A Response object that will be sent to correlated reverse proxy request.")
+
+          properties do
+            rig(
+              Schema.new do
+                properties do
+                  correlation(:string, "Correlation ID",
+                    required: true,
+                    example: "g2dkAA1ub25vZGVAbm9ob3N0AAADxwAAAAAA"
+                  )
+                end
+              end
+            )
+          end
+        end
+    }
+  end
+end
--- a/apps/rig_api/lib/rig_api/controllers/responses_controller.ex
+++ b/apps/rig_api/lib/rig_api/controllers/responses_controller.ex
@@ -5,6 +5,7 @@ defmodule RigApi.ResponsesController do
   use PhoenixSwagger
 
   alias Rig.Connection.Codec
+  alias RigCloudEvents.CloudEvent
 
   action_fallback(RigApi.FallbackController)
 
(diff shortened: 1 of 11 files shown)


kevinbader:
`http_internal` => what about `http_async`?

apps/rig_api/lib/rig_api/controllers/responses_controller.ex
```diff
@@ -0,0 +1,78 @@
...
+        |> put_status(:bad_request)
+        |> text("Failed to parse request body: #{inspect(err)}")
+    end
+  end
+
+  def swagger_definitions do
+    %{
+      Response:
+        swagger_schema do
+          title("Response object")
```
kevinbader: the "Response object" should actually be a CloudEvent

apps/rig_inbound_gateway/config/config.exs
```diff
@@ -94,7 +94,8 @@ config :rig, RigInboundGateway.ApiProxy.Router,
 config :rig, RigInboundGateway.ApiProxy.Handler.Http,
   cors: {:system, "CORS", "*"},
   kafka_response_timeout: {:system, :integer, "PROXY_KAFKA_RESPONSE_TIMEOUT", 5_000},
-  kinesis_response_timeout: {:system, :integer, "PROXY_KINESIS_RESPONSE_TIMEOUT", 5_000}
+  kinesis_response_timeout: {:system, :integer, "PROXY_KINESIS_RESPONSE_TIMEOUT", 5_000},
+  http_response_timeout: {:system, :integer, "PROXY_HTTP_RESPONSE_TIMEOUT", 5_000}
```
kevinbader: if you decide to go with `http_async`, this could be called `http_async_response_timeout` to differentiate it from the timeout used when waiting for the _synchronous_ response

mmacai:
> `http_internal` => what about `http_async`?

Might be, my idea was to avoid sync/async since it might be confusing for people.

apps/rig_tests/test/proxy/response_from/http_test.exs
```diff
@@ -1,15 +1,21 @@
 defmodule RigTests.Proxy.ResponseFrom.HttpTest do
   @moduledoc """
   If `response_from` is not set, the response is taken from the proxied request.
+  If `response_from` is set to http_internal, the response is taken from internal HTTP endpoint /v1/responses
```
kevinbader: similar to how we did it for the Kafka response_from target, I'd prefer having a dedicated test module for this feature

apps/rig_tests/test/proxy/response_from/http_test.exs
```diff
@@ -68,4 +74,76 @@ defmodule RigTests.Proxy.ResponseFrom.HttpTest do
...
+
+    route(endpoint_path, fn %{query: %{"correlation" => correlation_id}} ->
+      event =
+        Jason.encode!(%{
+          specversion: "0.2",
+          type: "rig.async-response",
+          source: "fake-service",
+          id: "1",
+          rig: %{correlation: correlation_id},
+          data: async_response
```
kevinbader: I think I'd do the async response outside the request handler, which would be the realistic use case and also make sure we're not simply forwarding the first response we got, but actually ignore the sync response

docs/rig-ops-guide.md
```diff
@@ -60,6 +60,7 @@ Variable&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
 `NODE_COOKIE` | Erlang cookie used in distributed mode, so nodes in cluster can communicate between each other. | nil
 `NODE_HOST` | Erlang hostname for given node, used to build Erlang long-name `rig@NODE_HOST`. This value is used by Erlang's distributed mode, so nodes can see each other. | nil
 `PROXY_CONFIG_FILE` | Configuration JSON file with initial API definition for API Proxy. Use this variable to pass either a path to a JSON file, or the JSON string itself. A path can be given in absolute or in relative form (e.g., `proxy/your_json_file.json`). If given in relative form, the working directory is one of RIG's `priv` dirs (e.g., `/opt/sites/rig/lib/rig_inbound_gateway-2.0.2/priv/` in a Docker container). | nil
+`PROXY_HTTP_RESPONSE_TIMEOUT` | In case an endpoint has `target` set to `http` and `response_from` set to `http_internal`, this is the maximum delay between an HTTP request and the corresponding HTTP response message. | 5000
```
kevinbader: maybe it's me but it doesn't come accross that it's not the _synchronous_ http response you're talking about here

kevinbader:
> > `http_internal` => what about `http_async`?
> 
> Might be, my idea was to avoid sync/async since it might be confusing for people.

The problem with `http_internal` is that "internal" refers to how we're used to call RIG's _internal_ API. But I don't think anybody else would relate. At least with async http we would make a statement that it's http, but not the sync response. Not ideal either, I know, but I still it's better than internal, to be honest.

apps/rig_tests/test/proxy/response_from/http_test.exs
mmacai: Can you elaborate how would you do it? Since, if you won't do the response in this handler it won't get pass `%HTTPoison.Response{status_code: res_status, body: res_body} = HTTPoison.get!(request_url)` due to timeout.

kevinbader: Indeed, that didn't make any sense ^_^ please ignore my comment.

PR #218, opened 2019-05-24
208 Integrity-check for the correlation ID

### LINKED ISSUE #208: Integrity-check for the correlation ID
## Background

The correlation ID is used for correlating an asynchronous backend response to the original request. The steps involved, in a nutshell:

1. A client connects to an endpoint that has `response_from` set to, e.g., `kafka`
1. RIG forwards the request, but before it does that, it adds a so-called "correlation ID" to the request.
1. Back-end services do their thing, while retaining that correlation ID. When they're ready to respond, they produce a message to Kafka that contains the correlation ID.
1. RIG decodes the correlation ID, **which is actually just the process ID of the still-open client connection**. With that `pid`, RIG can finally forward the response.

This works well, but is implemented in a pretty insecure way - the correlation ID doesn't include a signature of any kind, so RIG basically does an `eval` on incoming data => bad. Well, actually it's `binary_to_term`, but still it's quite ugly.

Here's the relevant code:
https://github.com/Accenture/reactive-interaction-gateway/blob/master/apps/rig/lib/rig/connection/codec.ex

## How to fix this

The correlation ID doesn't necessarily need to be encrypted, but RIG needs to be able to check the integrity of the string before interpreting it as a process ID. I'd probably go for a signature here, but the algorithm should be chosen by the one who implements this :)

pavestru:
## Description

- Add encryption and verification to pid serialization to achieve authenticity, integrity and confidentiality
- Encryption key read env var CODEC_SECRET_KEY

Closes #208 

## What to look out for

- do we need confidentiality?
- the new serialized string is these values concatenated: initialization_vector <> MAC <> ciphertext
- default value for secret_key is setup in apps/rig/config.exs
- no script for generating and setting CODEC_SECRET_KEY env var was implemented in this PR apart from setting the default value in config.exs
- need to implement a microbenchmark to measure impact on performance

### DIFF
--- a/apps/rig/config/config.exs
+++ b/apps/rig/config/config.exs
@@ -59,7 +59,7 @@ config :rig, Rig.EventStream.KafkaToHttp,
   group_id: {:system, "KAFKA_GROUP_ID", "rig"}
 
 config :rig, Rig.Connection.Codec,
-  secret_key: {:system, "CODEC_SECRET_KEY", "7tsf4Y6GTOfPY1iDo4PqZA=="}
+  codec_secret_key: {:system, "NODE_COOKIE", "7tsf4Y6GTOfPY1iDo4PqZA=="}
 
 config :porcelain, driver: Porcelain.Driver.Basic
 
--- a/apps/rig/config/config.exs
+++ b/apps/rig/config/config.exs
@@ -58,6 +58,9 @@ config :rig, Rig.EventStream.KafkaToHttp,
   targets: {:system, :list, "FIREHOSE_KAFKA_HTTP_TARGETS", []},
   group_id: {:system, "KAFKA_GROUP_ID", "rig"}
 
+config :rig, Rig.Connection.Codec,
+  secret_key: {:system, "CODEC_SECRET_KEY", "7tsf4Y6GTOfPY1iDo4PqZA=="}
+
 config :porcelain, driver: Porcelain.Driver.Basic
 
 import_config "#{Mix.env()}.exs"
--- a/apps/rig/config/config.exs
+++ b/apps/rig/config/config.exs
@@ -59,7 +59,8 @@ config :rig, Rig.EventStream.KafkaToHttp,
   group_id: {:system, "KAFKA_GROUP_ID", "rig"}
 
 config :rig, Rig.Connection.Codec,
-  codec_secret_key: {:system, "NODE_COOKIE", "7tsf4Y6GTOfPY1iDo4PqZA=="}
+  codec_secret_key: {:system, "NODE_COOKIE", nil},
+  codec_default_key: "7tsf4Y6GTOfPY1iDo4PqZA=="
 
 config :porcelain, driver: Porcelain.Driver.Basic
 
--- a/apps/rig/lib/rig/connection/codec.ex
+++ b/apps/rig/lib/rig/connection/codec.ex
@@ -1,18 +1,16 @@
 defmodule Rig.Connection.Codec do
   @moduledoc """
   Encode and decode a connection token, e.g., for correlation.
-
-  TODO: sign the token - using binary_to_term without signature verification is a threat.
   """
 
   @doc "Turn a pid into an url-encoded string."
   @spec serialize(pid) :: binary
   def serialize(pid) do
+    key = "7tsf4Y6GTOfPY1iDo4PqZA=="
     pid
     |> :erlang.term_to_binary()
+    |> encrypt(key)
     |> Base.url_encode64()
-
-    # TODO sign this
   end
 
   # ---
@@ -20,10 +18,11 @@ defmodule Rig.Connection.Codec do
   @doc "Convert a serialized string back into a pid."
   @spec deserialize(binary) :: {:ok, pid} | {:error, :not_base64 | :invalid_term}
   def deserialize(base64_encoded) do
-    # TODO check signature here
-
+    key = "7tsf4Y6GTOfPY1iDo4PqZA=="
     with {:ok, decoded_binary} <- decode64(base64_encoded) do
-      binary_to_term(decoded_binary)
+      decoded_binary
+      |> decrypt(key)
+      |> binary_to_term()
     end
   end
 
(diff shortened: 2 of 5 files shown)


kevinbader:
> Encryption key read env var CODEC_SECRET_KEY

There is already a secret key shared among nodes:

Variable&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Description | Default
-------------- | ----------- | -------
`NODE_COOKIE` | Erlang cookie used in distributed mode, so nodes in cluster can communicate between each other. | nil

What about re-using it here?

apps/rig/lib/rig/connection/codec.ex
```diff
@@ -1,29 +1,28 @@
...
   @moduledoc """
   Encode and decode a connection token, e.g., for correlation.
-
-  TODO: sign the token - using binary_to_term without signature verification is a threat.
   """
 
   @doc "Turn a pid into an url-encoded string."
   @spec serialize(pid) :: binary
   def serialize(pid) do
+    key = Confex.fetch_env!(:rig, Rig.Connection.Codec)[:secret_key]
```
kevinbader: I think I'd prefer `use Rig.Config, [:secret_key]` instead

pavestru: I've applied the change you suggested, but instead of `:secret_key` I've used name `:codec_secret_key`, because `secret_key` is used elsewhere as alias for `jwt_secret_key` as far as I understand. I'm using `NODE_COOKIE` env var.

kevinbader: well `:secret_key` in that context refers to the key used in the configuration _for this module_, which means there can't be a name clash here. That said, `:codec_secret_key` is fine, too.

apps/rig/config/config.exs
```diff
@@ -58,6 +58,9 @@ config :rig, Rig.EventStream.KafkaToHttp,
   targets: {:system, :list, "FIREHOSE_KAFKA_HTTP_TARGETS", []},
   group_id: {:system, "KAFKA_GROUP_ID", "rig"}
 
+config :rig, Rig.Connection.Codec,
+  codec_secret_key: {:system, "NODE_COOKIE", "7tsf4Y6GTOfPY1iDo4PqZA=="}
```
kevinbader: hm `mix compile` _should_ tell you that `NODE_COOKIE` cannot have more than one default value. Please use `nil` as the default, too. You can use this (or a random) string in the code below in case `NODE_COOKIE` is not set.

kevinbader: try running `MIX_ENV=prod mix compile`, which should give you the following change:

```bash
$ git diff
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: docs/rig-ops-guide.md
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ docs/rig-ops-guide.md:60 @ Variable&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
`KINESIS_OTP_JAR` | Path to the `OtpErlang.jar` file that contains the `JInterface` implementation. If left empty, RIG picks the file from its Erlang environment (Erlang must be compiled with Java support enabled). | nil
`KINESIS_STREAM` | The name of the Kinesis stream to consume. | "RIG-outbound"
`LOG_LEVEL` | Controls logging level for RIG, available values are: "debug", "info", "warn", "error". Production is using "warn" level. | :warn
`NODE_COOKIE` | Erlang cookie used in distributed mode, so nodes in cluster can communicate between each other. | nil
`NODE_COOKIE` | Erlang cookie used in distributed mode, so nodes in cluster can communicate between each other. | "7tsf4Y6GTOfPY1iDo4PqZA=="
`NODE_HOST` | Erlang hostname for given node, used to build Erlang long-name `rig@NODE_HOST`. This value is used by Erlang's distributed mode, so nodes can see each other. | nil
`PROXY_CONFIG_FILE` | Configuration JSON file with initial API definition for API Proxy. Use this variable to pass either a path to a JSON file, or the JSON string itself. A path can be given in absolute or in relative form (e.g., `proxy/your_json_file.json`). If given in relative form, the working directory is one of RIG's `priv` dirs (e.g., `/opt/sites/rig/lib/rig_inbound_gateway-2.0.2/priv/` in a Docker container). | nil
`PROXY_RECV_TIMEOUT` | Timeout used when receiving a response for a forwarded/proxied request. | 5000
```

kevinbader: just noticed that (1) you need to run mix compile with `MIX_ENV=prod` for the script to do anything and (2) the script ignores the `nil` default and uses the default you define here as the default for both settings \*sigh\*

So please just make sure both are set to `nil` :)

mmacai: @pavestru update on this?

pavestru: I've set the default value here to nil, added another default value in the same config and used it in code as follows:
```elixir
secret_key = conf.codec_secret_key || conf.codec_default_key
```
see [the commit](https://github.com/pavestru/reactive-interaction-gateway/commit/632b9969b5283c8b2f8eb6dcfe9cbe267b213ed7)

pavestru:
It is important that `NODE_COOKIE` is base64 decodable. Right now it is generated in the following way:
```
export NODE_COOKIE="$(cat /dev/urandom | env LC_CTYPE=C tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)"
```
Having 32 characters  'a-zA-Z0-9' is base64 decodable, so I think we're fine.

pavestru:
> I would also mention this change in `CHANGELOG.md`

I've done that.

PR #220, opened 2019-05-28
Support for receiving events via HTTP long polling

jakobwgnr:
Initial implementation of longpolling support.

Follow ups: #235, #236


apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
```diff
@@ -0,0 +1,80 @@
...
+    is_new_session =
+      if conn.req_cookies["connection_token"] === nil do
+        true
+      else
+        {:ok, session_pid} = Connection.Codec.deserialize(conn.req_cookies["connection_token"])
+        !Process.alive?(session_pid)
+      end
+
+    # No session set -> start new session
+    if(is_new_session) do
```
kevinbader: if/else expressions this long should go into a function that's named after what this is deciding and dispatch using pattern matching, like this:

```elixir
defp what_this_decides(is_new_session)
defp what_this_decides(true) do
  ...
end
defp what_this_decides(false) do
  ...
end
```

apps/rig_inbound_gateway/test/event_buffer_test.exs
```diff
@@ -0,0 +1,133 @@
+defmodule EventBufferTest do
+  use ExUnit.Case, async: true
+
+  alias RigInboundGatewayWeb.EventBuffer
+
+  test "New EventBuffer starts with no events" do
+    buffer = EventBuffer.new(5)
+    assert buffer.events === []
```
kevinbader: your tests rely on internals of `EventBuffer`. Also you write numbers into the buffer; either rename it to `Buffer` or write events into it. I'd prefer the latter - consider the following API:

```elixir
defp new_event(id), do: CloudEvent.parse("""
  {"specversion": "0.2", "id": "#{id}", "type": "test-event"}
""")

test "Adding events leaves the capacity unchanged and overwrites old events as soon as the buffer is filled up to its capacity.""" do
# We create a new buffer that holds at most 2 events:
buf = EventBuffer.new(2)
assert EventBuffer.capacity(buf) == 2
assert EventBuffer.all_events() == []

# We add a first event:
first_event = new_event("first")
buf = buf |> EventBuffer.add_event(first_event)
# The event is in the buffer:
assert EventBuffer.all_events() == [first_event]
# There is no more recent event yet:
assert EventBuffer.events_since("first") == []

# We add a second event to the buffer:
second_event = new_event("second")
buf = buf |> EventBuffer.add_event(second_event)
# Both events are now in the buffer:
assert EventBuffer.all_events() == [first_event, second_event]
# "second" is now more recent than "first", but there is no event newer than "second":
assert EventBuffer.events_since("first") == [second_event]
assert EventBuffer.events_since("second") == []

# Adding a third event should remove the first one:
third_event = new_event("third")
buf = buf |> EventBuffer.add_event(third_event)
# The first event is no longer in the buffer:
assert EventBuffer.all_events() == [second_event, third_event]
# "third" is now more recent than "second", but there is no event newer than "third":
assert EventBuffer.events_since("second") == [third_event]
assert EventBuffer.events_since("third") == []
# Since "first" is no longer in the buffer, _all_ events are newer than "first":
assert EventBuffer.events_since("first") == [second_event, third_event]

# The capacity hasn't changed during the modifications:
assert EventBuffer.capacity(buf) == 5
end
```

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
```diff
@@ -0,0 +1,185 @@
...
+          {write_pointer, read_pointer}
+        end
+      end
+
+    events = insert_or_update(events, write_pointer, entry)
+    write_pointer = write_pointer + 1
+
+    # When write_pointer is overtaking read_pointer on write - keep it synced 
+    read_pointer = if read_pointer === write_pointer, do: write_pointer + 1, else: read_pointer
+    read_pointer = if read_pointer === max_size, do: 0, else: read_pointer
```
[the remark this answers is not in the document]
jakobwgnr: Hmm - Not sure if I can apply a case statement here... 

Nevertheless the statement wasn't well written and therefore misleading. How about this?
```
    # When write_pointer is overtaking read_pointer on write - keep it synced    
    read_pointer =
      if read_pointer === write_pointer do
        if(read_pointer + 1 === max_size(do: 0, else: read_pointer + 1))
      else
        read_pointer
      end
```

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
```diff
@@ -0,0 +1,135 @@
...
+    case event_id do
+      "first_event" ->
+        new_events = get_new_events(events, 0, write_pointer, capacity)
+
+        case new_events do
+          [] ->
+            {:ok, [events: [], last_event_id: event_id]}
+
+          _ ->
+            {:ok, [events: new_events, last_event_id: CloudEvent.id!(Enum.at(new_events, -1))]}
```
mmacai: Would be good to have it in a more Elixir way:

```
last_event_id =
  new_events
  |> Enum.at(-1)
  |> CloudEvent.id!()
```

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
```diff
@@ -0,0 +1,135 @@
...
+            read_pointer = rem(event_index + 1, capacity)
+            new_events = get_new_events(events, read_pointer, write_pointer, capacity)
+
+            case new_events do
+              [] ->
+                {:ok, [events: [], last_event_id: event_id]}
+
+              _ ->
+                {:ok,
+                 [events: new_events, last_event_id: CloudEvent.id!(Enum.at(new_events, -1))]}
```
mmacai: Same as above.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
```diff
@@ -0,0 +1,135 @@
...
+        case Enum.find_index(events, fn x -> CloudEvent.id!(x) == event_id end) do
+          nil ->
+            next_event = Enum.at(events, 0)
+            {:no_such_event, [not_found_id: event_id, last_event_id: CloudEvent.id!(next_event)]}
+
+          event_index ->
+            read_pointer = rem(event_index + 1, capacity)
+            new_events = get_new_events(events, read_pointer, write_pointer, capacity)
+
+            case new_events do
```
mmacai: Perhaps this (and also the same `case` above) could be extracted to function with pattern matching and thus avoid deeply nested `case`s.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
```diff
@@ -0,0 +1,226 @@
...
+    end
+  end
+
+  # ---
+  # Waits for events and replies them if available
+  @impl true
+  def handle_call({:recv_events, last_event_id, _}, from, state) do
+    state = %{
+      state
+      | session_valid_until: DateTime.add(DateTime.utc_now(), @session_timout_ms, :millisecond)
```
mmacai: Getting here:

```
warning: function DateTime.add/3 is undefined or private
```

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
```diff
@@ -0,0 +1,226 @@
...
+
+  require Logger
+
+  alias Rig.EventFilter
+  alias RigCloudEvents.CloudEvent
+  alias RigInboundGatewayWeb.EventBuffer
+  alias RIG.Subscriptions
+  alias RigInboundGateway.Events
+
+  @window_size_ms 500
```
mmacai: Imo some of these values should be configurable via env vars. Question is, which might be interesting -- to not end up with crazy amount of configurations.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
```diff
@@ -0,0 +1,96 @@
...
+    conn = conn |> fetch_cookies
+
+    conn.req_cookies["connection_token"] |> is_new_session? |> process_request(conn)
+  end
+
+  # -----
+  # Helpers
+  # -----
+
+  # validates if a connection_token was given. If yes, it validates if corresponding session processes is still alive, ignoring invalid/timed out cookies
```
mmacai: processes are still alive

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
```diff
@@ -0,0 +1,96 @@
...
+      |> with_allow_origin()
+      |> put_resp_cookie("connection_token", session_pid |> Connection.Codec.serialize())
+      |> put_resp_cookie("last_event_id", Jason.encode!("first_event"))
+      |> put_resp_header("content-type", "application/json; charset=utf-8")
+      |> put_resp_header("cache-control", "no-cache")
+      |> send_resp(200, Jason.encode!("ok"))
+    else
+      {:error, message} ->
+        conn
+        |> send_resp(
```
mmacai: I would stick here with a pattern we are using in other places, such as `apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/subscription_controller.ex:94`.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
```diff
@@ -0,0 +1,96 @@
...
+      )
+
+    conn
+    |> with_allow_origin()
+    |> put_resp_cookie("connection_token", session_pid |> Connection.Codec.serialize())
+    |> put_resp_cookie("last_event_id", Jason.encode!(response[:last_event_id] || "first_event"))
+    |> put_resp_header("content-type", "application/json; charset=utf-8")
+    |> put_resp_header("cache-control", "no-cache")
+    |> send_resp(
+      200,
```
mmacai: Same here.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
```diff
@@ -0,0 +1,96 @@
...
+        |> send_resp(
+          400,
+          "PARSE ERROR: #{inspect(message)}; SENT PARAMS: #{inspect(conn.query_params)})"
+        )
+    end
+  end
+
+  # reconnect to existing session
+  defp process_request(false, conn) do
+    {:ok, session_pid} = Connection.Codec.deserialize(conn.req_cookies["connection_token"])
```
mmacai: We got the `session_pid` already in `is_new_session?`, can we perhaps re-use it?

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
```diff
@@ -0,0 +1,96 @@
...
+          _ -> false
+        end
+    end
+  end
+
+  # ---
+  defp process_request(is_new_session, conn)
+
+  # starts a new session
+  defp process_request(true, conn) do
```
mmacai: Looks like a lot of repetitive code, what do you think about updating it to `with` statement, such as in `apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/subscription_controller.ex`? Flow there is pretty similar.

examples/README.md
```diff
@@ -1,12 +1,13 @@
...
+Following examples are show casing using of live updates via Server-sent events (SSE), Longpolling and Websocket (WS). Shows how to use public as well as private delivery.
 
-## SSE
+# SSE
+## Example 1: SSE
 
 This example shows simplest scenario when all messages will arrive to UI.
 
-> examples/sse-demo.html
+> examples/sse-1-demo.html
```
mmacai: What is the intention here with the changes?

examples/README.md
```diff
@@ -149,3 +153,80 @@ This time we are not calling subscription call, but they are automatically creat
 
 - Fill `john.doe` (this is set in JWT) to first input and some message to second input => update should be displayed below after few moments
 - Fill any other name to first input => this time no update should arrive
+
+# Longpolling
```
mmacai: Would be good to include examples as soon as possible -- to properly test all important cases.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
```diff
@@ -0,0 +1,135 @@
...
+       when read_pointer == write_pointer do
+    []
+  end
+
+  @spec get_between(List.t(), number, number) :: List.t()
+  defp get_between(_list, _from_index, to_index) when to_index < 0 do
+    []
+  end
+
+  @spec get_between(List.t(), number, number) :: List.t()
```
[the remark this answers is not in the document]
jakobwgnr: what do you mean, don't fully understand :)

examples/README.md
jakobwgnr: I was a bit struggling to find the right file to the examples so I numbered them according the the Example-no. :) what do you think?

examples/README.md
jakobwgnr: yes - unfortunately i didn't find time yet to do so...

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
jakobwgnr: I had it that way, but kevin pointed out that doing it this way increases the readability of the main function so i changed it to that

apps/rig_inbound_gateway/test/event_buffer_test.exs
```diff
@@ -0,0 +1,81 @@
+defmodule EventBufferTest do
```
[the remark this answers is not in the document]
jakobwgnr: fixed

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
```diff
@@ -0,0 +1,135 @@
...
+  """
+  @spec all_events(EventBuffer.t()) :: List.t()
+  def all_events(buffer) do
+    buffer.events
+  end
+
+  @doc """
+  writes a value to the given EventBuffer, overwrites the oldest value if EventBuffer is full
+  """
+  @spec add_event(EventBuffer.t(), CloudEvent.t()) :: EventBuffer.t()
```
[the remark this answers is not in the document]
jakobwgnr: hmm - strange I don't get this... which mix task do you use here?

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
jakobwgnr: @kevinbader your opinion?

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
jakobwgnr: hmm - can't find a good way how.. thought that ?-methods should only return true/false?

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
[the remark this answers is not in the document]
mmacai: `rm -rf _build && mix test`

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/event_buffer.ex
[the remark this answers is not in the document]
mmacai: I mean to use just single spec:

```
@spec fn ...
defp fn() do ...

defp fn() do ...

...
```

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/v1/longpolling_controller.ex
mmacai: `is_new_session ` could return tuple with `{boolean, pid}`, but you have correct point, perhaps better to keep it like this.

examples/README.md
mmacai: Ok, I guess that makes sense.

mmacai:
Hmm, still getting the same error when running `mix test` :/ (tried also `rm -rf _build`). @kevinbader can you give it a try?

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
mmacai: Imo, at least `event_buffer_size ` and `window_size_ms` should be configurable via env vars. When I think about it as a user/consumer, I would probably change those 2 according to my needs.

kevinbader: I agree with `event_buffer_size` but not with `window_size_ms`. The latter is an internal thing, basically the polling interval for checking whether there are events available. Btw, perhaps it would be a good idea to rename that accordingly - it's not related to [what I would consider a window](https://en.wikipedia.org/wiki/Sliding_window_protocol) in this context.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
```diff
@@ -0,0 +1,226 @@
...
+
+      {:no_such_event, [not_found_id: _not_found_id, last_event_id: last_event_id]} ->
+        # the event_id provided by the client was outdated - we send the newest event id
+        {:reply, %{last_event_id: last_event_id, events: [], status: :no_such_event}, state}
+    end
+  end
+
+  # Waits for events and replies them if available (same as above, but as handle_info)
+  @impl true
+  def handle_info({:recv_events, from, last_event_id, tries}, state) do
```
kevinbader: shouldn't this function also update `session_valid_until`?

apps/rig_inbound_gateway/test/event_buffer_test.exs
```diff
@@ -0,0 +1,82 @@
...
+    first_event = new_event("first")
+    buffer = buffer |> EventBuffer.add_event(first_event)
+    # The event is in the buffer:
+    assert buffer |> EventBuffer.all_events() == [first_event]
+    # There is no more recent event yet:
+    {:ok, [events: events, last_event_id: last_event_id]} =
+      buffer |> EventBuffer.events_since("first")
+
+    assert events == []
+    assert last_event_id == "first"
```
kevinbader: it's not wrong or anything, but we usually write a test like this in one line:

```elixir
assert {:ok, [events: [], last_event_id: "first"]} = EventBuffer.events_since(buffer, "first")
```

Also, imho using a pipeline operator for a single function call makes the code a little bit harder to read.

apps/rig_inbound_gateway/test/event_buffer_test.exs
```diff
@@ -0,0 +1,82 @@
...
+
+    # We add a second event to the buffer:
+    second_event = new_event("second")
+    buffer = buffer |> EventBuffer.add_event(second_event)
+    # Both events are now in the buffer:
+    assert buffer |> EventBuffer.all_events() == [first_event, second_event]
```
kevinbader: imho, using a pipeline operator for a single function call makes the code a little bit harder to read.

kevinbader:
@jakobwgnr I made the EventBuffer module a bit simpler - please take a look at [the PR in your fork](https://github.com/jakobwgnr/reactive-interaction-gateway/pull/1).

kevinbader:
> Hmm, still getting the same error when running `mix test` :/ (tried also `rm -rf _build`). @kevinbader can you give it a try?

hm seems to work for me (running Elixir 1.8.2) - I don't see any errors besides the usual Ranch protocol crash_reason=undef ones.

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
```diff
@@ -0,0 +1,232 @@
...
+  @session_timeout_validation_interval_ms 60_000
+  @session_timeout_ms 3_600_000
+
+  # ---
+  def start(query_params, opts \\ []) do
+    GenServer.start(__MODULE__, %{query_params: query_params}, opts)
+  end
+
+  def recv_events(server, last_event_id) do
+    GenServer.call(server, {:recv_events, last_event_id || "first_event", 0}, 20_000)
```
kevinbader: if you accept my changes to EventBuffer, this hardcoded "first_event" should be removed

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
jakobwgnr: yep

jakobwgnr:
@kevinbader @mmacai 
Fixed latest review including merge of the PR. 
Manual Test worked as expected
automated test worked as expected
no credo issues

apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/session.ex
kevinbader: resolving in favor of #236

PR #224, opened 2019-06-05
Turn :refresh_subscriptions call into cast

kevinbader:
Refreshing subscriptions doesn't trigger/require a response, so cast is
appropriate. It also helps with preventing unnecessary GenServer timeouts in
case the server is under load.


mmacai:
Seems like tests are failing, can you check it?

PR #228, opened 2019-06-27
227 cypress

### LINKED ISSUE #227: Functional tests for examples
## Summary and context

Right now we have to manually check every example when new release or any change happens. Which is ... quite annoying and possibly error-prone. Therefore, it would be great to have an automated way how to do it and ideally also part of the Travis pipeline.

To keep our codebase "simple" we may use https://www.cypress.io/, since we are already using JavaScript.

What should be done:

- [x] cover simple HTML examples
- [x] cover channels example
- [x] add missing simple HTML example to showcase how to do SSE/WS connection and subscription within the same HTTP call
- [x] make it part of the Travis pipeline

Additionally, every new example should automatically include also test case for this.

mmacai:
## Description

- added functional tests with Cypress for examples
- integrated into Travis pipeline
- added missing example for creating sse/ws connection and subscriptions within a same call
- updated `.dockerignore`
- updated js dependencies across the board - this should hopefully solve all (or most of the) security warnings

Closes #227

## What to look out for

Dear reviewer, I want you to

- check tests if they make sense
- check the entire setup around them, perhaps there is some better way how to instrument tests with different RIG configurations
- check Travis pipeline logs

## Process

Have fun!


CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-<!-- ### Added -->
+### Added
+
+- [Examples] Added functional tests for all examples using Cypress. [#227](https://github.com/Accenture/reactive-interaction-gateway/issues/227)
```
kevinbader: I must admit I'm not a huge fan of those `[]` categories. The Changelog target _users_ and they pretty much have no clue how we organize our source code (and rightfully so). Also, even for us developers, `[Examples]` doesn't add much value since I need to read the rest of the line anyway. In this case, the sentence on its own already tells me that it's about the examples; in other cases, one might set the scene with words rather than with text. For example, I think "When forwarding outbound HTTP responses, RIG makes sure the HTTP version of the response matches what the clients expects." better than the more technical "[Proxy] HTTP procotol version-conversion for outbound responses.". Does that make sense? :)

examples/tests/cypress/integration/channels.spec.js
```diff
@@ -0,0 +1,31 @@
+describe('Channels', () => {
+  ['sse', 'ws'].forEach(type => {
+    it(`${type} Public events`, () => {
```
kevinbader: "it sse public events" is not the most descriptive test name. What does this test? Does `connectAndSendEvents` also check that the message sent has been received afterwards? If yes, shouldn't the function name hint on that? If not, then this test only tests that sending something "works".. probably.. but I don't see any assertions ;)

examples/tests/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "cypress:open:jwt": "cypress open --spec 'cypress/integration/subscriptions-jwt.spec.js'",
+    "cypress:open:channels": "cypress open --spec 'cypress/integration/channels.spec.js'",
+    "cypress:run": "cypress run",
+    "cypress:run:noauth": "NO_COLOR=1 cypress run --spec 'cypress/integration/subscriptions-no-auth.spec.js'",
+    "cypress:run:jwt": "NO_COLOR=1 cypress run --spec 'cypress/integration/subscriptions-jwt.spec.js'",
+    "cypress:run:channels": "NO_COLOR=1 cypress run --spec 'cypress/integration/channels.spec.js'",
+    "cypress:run:all": "NO_COLOR=1 ./run_all_tests.sh"
+  },
+  "author": "",
+  "license": "ISC",
```
kevinbader: I guess we should use the same license everywhere

examples/tests/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "cypress:open:noauth": "cypress open --spec 'cypress/integration/subscriptions-no-auth.spec.js'",
+    "cypress:open:jwt": "cypress open --spec 'cypress/integration/subscriptions-jwt.spec.js'",
+    "cypress:open:channels": "cypress open --spec 'cypress/integration/channels.spec.js'",
+    "cypress:run": "cypress run",
+    "cypress:run:noauth": "NO_COLOR=1 cypress run --spec 'cypress/integration/subscriptions-no-auth.spec.js'",
+    "cypress:run:jwt": "NO_COLOR=1 cypress run --spec 'cypress/integration/subscriptions-jwt.spec.js'",
+    "cypress:run:channels": "NO_COLOR=1 cypress run --spec 'cypress/integration/channels.spec.js'",
+    "cypress:run:all": "NO_COLOR=1 ./run_all_tests.sh"
+  },
+  "author": "",
```
kevinbader: Accenture or your name or both

examples/tests/run_all_tests.sh
```diff
@@ -0,0 +1,45 @@
...
+
+section_header "Running functional test suite for Examples with no auth"
+docker rm -f rig || true
+docker run -d --name rig \
+-e EXTRACTORS='{"greeting":{"name":{"stable_field_index":1,"event":{"json_pointer":"/data/name"}}},"greeting.jwt":{"name":{"stable_field_index":1,"jwt":{"json_pointer":"/username"},"event":{"json_pointer":"/data/name"}}},"nope":{"fullname":{"stable_field_index":1,"jwt":{"json_pointer":"/fullname"},"event":{"json_pointer":"/data/fullname"}}},"example":{"email":{"stable_field_index":1,"event":{"json_pointer":"/data/email"}}}}' \
+-e JWT_SECRET_KEY=secret \
+-p 4000:4000 \
+accenture/reactive-interaction-gateway
+
+npm run cypress:run:noauth
```
kevinbader: it would be nice (not mandatory) to fail right at the beginning in case cypress is not there cause `npm install` has not yet been invoked

CHANGELOG.md
mmacai: Ye, me neither ... I was always thinking about `[]` as a way how to filter all changes for particular module/sub-app/... but we are not following any pattern/style for that really ... what do you propose? should we get rid of them?

examples/tests/cypress/integration/channels.spec.js
mmacai: ye, I guess it's poor naming, I'll change it -- you can find assertions in `commands.js`

mmacai: Updated structure and naming in general, should be easier to follow now, let me know if it's enough or you see some places for improvement.

CHANGELOG.md
kevinbader: I think we either have to use very few (consistently) and define them clearly, or we should omit them altogether. I lean towards the latter.

examples/tests/cypress/integration/channels.spec.js
kevinbader: 👍

kevinbader: very cool - way more readable now!

kevinbader:
> check the entire setup around them, perhaps there is some better way how to instrument tests with different RIG configurations

anything in particular that makes you uncomfortable about the setup? if it'd make sense to change certain env vars during the test run, perhaps we could fiddle something together using a remote shell and system.set_env. or write a dedicated confex adapter for this.

mmacai:
> > check the entire setup around them, perhaps there is some better way how to instrument tests with different RIG configurations
> 
> anything in particular that makes you uncomfortable about the setup? if it'd make sense to change certain env vars during the test run, perhaps we could fiddle something together using a remote shell and system.set_env. or write a dedicated confex adapter for this.

Not really, I'm just every time curious if I didn't miss simpler solution when doing shell scripts.

mmacai:
@kevinbader is there anything else I should change? I guess the cleanup of changelog will happen in different PR.

PR #231, opened 2019-07-12
WIP: 226 joken2 dep upgrade

arana3:
WIP PR for updating RIG's Joken v1 dep to v2 dep.

Due to v2 breaking API changes, code modifications must be made across core.


apps/rig_auth/lib/rig_auth/jwt/TokenUtils.ex
```diff
@@ -0,0 +1,52 @@
+defmodule RigAuth.Jwt.TokenUtils do
+  use Joken.Config, default_signer: nil # no default signer. Do not set as that will break Jwks retrieval
+
+  use Rig.Config, [:secret_key, :rotation_flag, :jwks_endpoint]
+
+  @conf_var config()
```
kevinbader: by doing this you effectively "freeze" the configuration into the compiled module - as long as the module is not recompiled, the configuration cannot be changed. Considering that Confex specifically allows to _reload_ configuration at runtime (e.g., by changing the environment variables with the default Confex adapter), I think it'd be better not call `config()` where you need it instead.

apps/rig_auth/lib/rig_auth/jwt/TokenUtils.ex
```diff
@@ -0,0 +1,52 @@
...
+
+  def config_retrieval do
+    @conf_var
+  end
+
+  if conf_var.rotation_flag do
```
kevinbader: similar to above, this code is executed when the module is compiled, which might not be what you want

mmacai:
@arana3 any chance you finish this one?

arana3:
> @arana3 any chance you finish this one?

Sadly, I have not. Should we mark as in-complete and come back to it when it is a priority?

mmacai:
@arana3 Ye, let's close it for now and re-open/do new PR once it gets some updates.

PR #239, opened 2019-07-31
Upgrade to Elixir 1.9.1

kevinbader:
## Description

Upgrade the Elixir and Erlang versions for source code and Docker images.

## What to look out for

From now on, releases are no longer built using `mix release`; use `mix distillery.release` instead.

## Process

Have fun!


.travis.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   include:
     - stage: unittest
       language: elixir
-      elixir: "1.7"
+      elixir: "1.9"
       otp_release: "21.0"
```
mmacai: Shouldn't be otp release bumped up as well?

.travis.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         - cd website && yarn install && GIT_USER="${GH_NAME}" yarn run publish-gh-pages
     - stage: sourcedocs
       language: elixir
-      elixir: "1.7"
+      elixir: "1.9"
       otp_release: "21.0"
```
mmacai: Same here.

PR #247, opened 2019-08-09
Fix missing request body for external subscription check

kevinbader:
## Description

When using an external service to authorize subscriptions, this makes sure the subscription body is actually passed in the request body.

Bug report on Slack:

> Piotr Kłoskowski 6:25 PM
looks like subscription parameters aren’t being passed to body:
request: %HTTPoison.Request{body: "", headers: [{"content-type", "application/json; charset=utf-8"}, {"authorization", "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6Im1pa2UiLCJpYXQiOjE1NjUyODEyOTYsImV4cCI6MTU2NTI4NDg5Nn0.J_a8p0MMgrSQ6n-oJ29BZ4FhR429DAxAiO8FUpPNifM"}], method: :post, options: [], params: %{}, url: "http://validator/"}, request_url: "http://validator/", status_code: 500}
Validator:
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

Note that there is no Changelog entry as this feature hasn't been released yet.

### DIFF
--- a/apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/connection_init.ex
+++ b/apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/connection_init.ex
@@ -29,9 +29,7 @@ defmodule RigInboundGatewayWeb.ConnectionInit do
         subscription_refresh_interval_ms
       ) do
     Logger.debug(fn ->
-      "new #{conn_type} connection (pid=#{inspect(self())}, params=#{
-        inspect(request.query_params)
-      })"
+      "new #{conn_type} connection (pid=#{inspect(self())}, params=#{inspect(request)})"
     end)
 
     jwt =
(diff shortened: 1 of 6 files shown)


apps/rig_inbound_gateway/lib/rig_inbound_gateway_web/connection_init.ex
```diff
@@ -64,4 +62,27 @@ defmodule RigInboundGatewayWeb.ConnectionInit do
         on_error.("Subscription denied (not authorized).")
     end
   end
+
+  # ---
+
+  @spec subscriptions_query_param_to_body(map) :: {:ok, String.t()} | {:error, String.t()}
+  def subscriptions_query_param_to_body(query_params)
```
mmacai: Should be perhaps `subscriptions_query_params_to_body`? Since it's accepting multiple params not just single one.

Issue #157, opened 2019-02-05
RIG specific prometheus metrics


jakobwgnr:
This issue is supposted to cover implementation of RIG specifc metrics.

1. implement "setup" of metrics within the instrumenter modules under `rig_metrics/lib/rig_metrics/*` Those are already implemented for the planned metrics, but commented out right now to avoid exposing not working metrics
2. implement metrics triggers for the varying use cases in the other otp apps (might differ per metrics). For this a dependency to rig_metrics will be required in the other otp apps

**Metrics planned as of latest discussion:**

RigProxyInstrumenter

* rig_current_session_count --> As Prometheus Gauge
* rig_current_subscription_count --> As Prometheus Gauge
* rig_current_open_proxy_connection_count --> As Prometheus Gauge
* rig_proxy_requests_total --> As Prometheus Counter

RigEventhubInstrumenter

* rig_processed_events_total --> As Prometheus counter + Label Node (e.g. Kafka, etc)
* rig_forwarded_events_total --> As Prometheus counter + Label Node (e.g. Kafka, etc)
* rig_dropped_events_total --> As Prometheus counter + Label Node (e.g. Kafka, etc)

RigControlInstrumenter

* rig_current_session_blacklisted_count --> As Prometheus Gauge
* rig_current_routes_configured_count --> As Prometheus Gauge

**Further Metric ideas:**
- Messages_total
- Messages_Error
- Latency
- Throughput
- etc.

PR #260, opened 2019-09-04
Refine session related code: rename it, add tests, fix bugs, update docs

kevinbader:
## Description

- Got rid of `RigAuth` and moved the `Session` module as well as the authorization-check modules under the `RIG` namespace.
- Fix: A connection is now associated to its session right after the connection is established, given the request carries a JWT in its authorization header.
- Added unit tests for the session module.

## What to look out for

Dear reviewer, I want you to

- help to find bugs.

Have fun!


lib/rig/distributed_set.ex
```diff
@@ -258,10 +258,11 @@ defmodule Rig.DistributedSet do
 
     n_deleted = :ets.select_delete(ets_table, match_spec)
 
-    if n_deleted > 0 do
-      Logger.debug(fn -> "Removed #{n_deleted} expired records" end)
-      RigMetrics.ControlInstrumenter.delete_blacklisted_session(n_deleted)
```
mmacai: Isn't this used in Prometheus?

mmacai:
Running smoke tests and getting logs below, can you confirm?

```
╓─────[ Running tests, plaintext Kafka connection ]─────╖

Starting zookeeper ... done
Starting rest-api  ... done
Starting kafka     ... done
08:09:37.074 module=Rig.Config [error] Could not resolve HTTPS_CERTFILE: "cert/selfsigned.pem"
08:09:37.092 module=Rig.Config [error] Could not resolve HTTPS_KEYFILE: "cert/selfsigned_key.des3.pem"
```

lib/rig/distributed_set.ex
kevinbader: not sure... it didn't work that way. Looking at the [instrumenter module](https://github.com/Accenture/reactive-interaction-gateway/blob/master/lib/rig_metrics/control_instrumenter.ex), it seemed like this was still work-in-progress, so I've gathered it's better to remove it and add the feature later for real. @jakobwgnr what do you think?

kevinbader: should be addressed in #157

PR #261, opened 2019-09-08
Add blacklist i9n test and fix some stuff along the way

kevinbader:
## Description

Added:

- When terminating an SSE connection after its associated session has
been blacklisted, RIG now sends out a `rig.session_killed` event before
closing the socket.
- New API for querying and updating the session blacklist: `/v2/session-blacklist`, which introduces the following breaking changes:
  - When a session has been added to the session blacklist successfully, the endpoint now uses the correct HTTP status code "201 Created" instead of "200 Ok".
  - When using the API to blacklist a session, the `validityInSeconds` should now be passed as an integer value (using a string still works though).

Removed:

- Removed the `JWT_BLACKLIST_DEFAULT_EXPIRY_HOURS` environment variable
([deprecated since 2.0.0-beta.2]).

Security:

- A connection is now associated to its session right after the
connection is established, given the request carries a JWT in its
authorization header. Previously, this was only done by the
subscriptions endpoint, which could cause a connection to remain active
even after blacklisting its authorization token.

[deprecated since 2.0.0-beta.2]: https://github.com/Accenture/reactive-interaction-gateway/commit/f974533455aa3ebc550ee95bf291585925a406d5

On a side note, I've also improved the logging a bit: in `debug` level, there haven't been any logs in case the `on_error` callback was invoked in the `ConnectionInit` module; now RIG will tell us in case something went wrong, and even output all the outgoing messages, at least for SSE.

## What to look out for

Dear reviewer, I want you to

I know that we still don't have anything on the blacklist feature on the website, but let's resolve this in a follow-up PR.


test/blacklist_test.exs
```diff
@@ -0,0 +1,95 @@
...
+  defp blacklisted?(jti) do
+    case HTTPoison.get("#{@rig_api_url}/v2/session-blacklist/#{URI.encode(jti)}") do
+      {:ok, %HTTPoison.Response{status_code: 200}} -> true
+      {:ok, %HTTPoison.Response{status_code: 404}} -> false
+    end
+  end
+
+  # ---
+
+  defp new_jwt(claims) do
```
mmacai: Any reason to have this private function and not use `JWT.encode(...)` directly in test cases?

kevinbader: good catch - I've used a different `jwt_conf` in there previously, but this way this `defp` no longer makes sense indeed => removed

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added longpolling as new connection type [#217](https://github.com/Accenture/reactive-interaction-gateway/issues/217)
+- When terminating an SSE connection after its associated session has been blacklisted, RIG now sends out a `rig.session_killed` event before closing the socket.
```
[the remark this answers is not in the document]
kevinbader: Added a description for the WebSocket behavior.

PR #262, opened 2019-09-17
WIP: 154 Metadata endpoint stub

Azer0s:
## Description

This adds the stub for the metadata endpoint for the new Session API. The whole idea here is that the frontend can send metadata to RIG. One can configure indices, use values from the JWT as metadata and set aliases for the JWT values. The RIG instance that receives the metadata from a client will extract data from the JWT, check the metadata set contains all indices and forward all indices to all other RIG instances.

### Example 1 - Simple index

Say, you have your RIG instance configured so you have an index on the field `userid`. This means the config would look like so: 

```elixir
config :rig, RigInboundGatewayWeb.V1.MetadataController,
  jwt_fields: %{},
  indexed_metadata: ["userid"]
```

And you send a metadata JSON that looks like this:

```js
{
    "metadata":
    { 
        "userid": "9ab1bff2-a8d8-455c-b48a-50145d7d8e30",
        "locale": "de-AT",
        "timezone": "GMT+2"
    }
}
```

RIG would distribute `userid` and the connection token to all other RIG instances to tell them where the metadata for these keys are.

### Example 2 - Index on extracted value

One can also extract values from the JWT. Say, we extract the `userid` value from the `sub` field in our JWT with a configuration looking like this:

```elixir
config :rig, RigInboundGatewayWeb.V1.MetadataController,
  jwt_fields: %{"userid" => "sub"},
  indexed_metadata: ["userid"]
```

One could send the following JWT: 

```jwt
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI5YWIxYmZmMi1hOGQ4LTQ1NWMtYjQ4YS01MDE0NWQ3ZDhlMzAiLCJuYW1lIjoiSm9obiBEb2UiLCJpYXQiOjE1Njg3MTMzNjEsImV4cCI6NDEwMzI1ODE0M30.kjiR7kFyOEeMJaY1zPCctut39eEWmKswUCNZdK5Q3-w
```

Along with the following JSON:

```js
{
    "metadata":
    { 
        "locale": "de-AT",
        "timezone": "GMT+2"
    }
}
```

## What to look out for

As this is just a WIP PR for code review please:
- check that everything works on your machine.
- test the endpoint (as described under "Dirty testing with JWT" and "Dirty testing without JWT" in lib/rig_inbound_gateway_web/v1/metadata_controller.ex).

### DIFF
--- a/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
+++ b/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
@@ -70,18 +70,16 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
          {:parse, %{"metadata" => metadata}} <- {:parse, json},
          {:idx, true} <- {:idx, contains_idx(metadata)} do
 
-      IO.inspect metadata
+      IO.puts inspect(metadata)
 
       send_resp(conn, :no_content, "")
     else
       {:idx, false} ->
-        
         send_resp(conn, :bad_request, """
         Metadata doesn't contain indexed fields.
         """)
 
       {:parse, json} ->
-
         send_resp(conn, :bad_request, """
         Expected field "metadata" is not present.
 
@@ -90,7 +88,6 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
         """)
 
       error ->
-
         send_resp(conn, :bad_request, """
         Expected JSON encoded body.
 
--- a/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
+++ b/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
@@ -14,7 +14,7 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
 
   require Logger
 
-  @spec set_metadata( conn :: Plug.Conn.t(), params :: map) :: Plug.Conn.t()
+  @spec set_metadata(conn :: Plug.Conn.t(), params :: map) :: Plug.Conn.t()
   def set_metadata(
       %{method: "PUT"} = conn,
       %{
--- a/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
+++ b/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
@@ -1,6 +1,6 @@
 defmodule RigInboundGatewayWeb.V1.MetadataController do
   @moduledoc """
-  Handles inbound metadata sets and the indexing and distribution of metadata 
+  Handles inbound metadata sets and the indexing and distribution of metadata
   """
   use RigInboundGatewayWeb, :controller
 
@@ -21,7 +21,7 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
   Sets metadata KV pairs for an existing connection, replacing previous metadata sets.
 
   There may be indexed metadata pairs defined. If a metadata set does not contain these indexed keys, it cannot be accepted.
-  
+
   ## Example
 
   Metadata set that contains a user ID, locale and timezone:
@@ -34,7 +34,7 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
         }
       }
 
-  In this example, the field "userid" might be indexed. In this case, the connection token would get associated with the value of the "userid" field (it will be possible to automatically take the user id from the JWT token (so that it cannot be changed by any other way) and index that; this will be serverside configuration). 
+  In this example, the field "userid" might be indexed. In this case, the connection token would get associated with the value of the "userid" field (it will be possible to automatically take the user id from the JWT token (so that it cannot be changed by any other way) and index that; this will be serverside configuration).
   In addition to that, the value of the "userid", "locale" and "timezone" fields would get associated to the connection token.
 
   To sum it up: the connection token is indexed by default, while all other fields can be indexed by configuration.
@@ -42,17 +42,17 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
   So the association would look like so:
 
       userid -> connection token
-      
+
       conection token ->
         userid
         locale
         timezone
 
-  All indexed fields are then being propagated to all other instances so they know where to find the corresponding metadata.
+  All indexed fields are then being propagated to all other instances, so they know where to find the corresponding metadata.
 
   ## Example
 
-  A user connects to RIG 1 and sends the metadata fields from above to RIG 1. RIG 1 would propagate the connection token and the userid along with the RIG instance ID to all other RIG instances.
+  A user connects to RIG 1 and sends the metadata fields from above to RIG 1. RIG 1 would propagate the connection token and the userid along with the RIG instance ID to all other RIG instances (via gen_tcp, preferably).
   If someone were to access the metadata of a user from RIG 3, RIG 3 would know where to find the metadata, ask RIG 1 for it and return the metadata set.
 
   """
@@ -64,7 +64,7 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
       }
   ) do
     IO.puts "REACHED SET METADATA"
-    IO.puts(BodyReader.read_full_body(conn) |> elem(1))
+    IO.puts(inspect(BodyReader.read_full_body(conn)))
 
     send_resp(conn, :no_content, "")
   end
--- a/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
+++ b/lib/rig_inbound_gateway_web/v1/metadata_controller.ex
@@ -3,6 +3,7 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
   Handles inbound metadata sets and the indexing and distribution of metadata
   """
   use RigInboundGatewayWeb, :controller
+  use Rig.Config, [:indexed_metadata]
 
   alias Result
 
@@ -98,8 +99,9 @@ defmodule RigInboundGatewayWeb.V1.MetadataController do
   end
 
   def contains_idx(metadata) do
-    # TODO: Actually check this by configuration
-    Enum.any?(metadata, fn x -> (x |> elem(0)) === "userid" end)
+    conf = config()
+    metadata_keys = Enum.map(metadata, fn x -> x |> elem(0) end)
+    Enum.all?(conf.indexed_metadata, fn x -> Enum.member?(metadata_keys, x) end)
   end
 
 end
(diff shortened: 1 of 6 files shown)


lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+
+      with {:ok, auth_tokens} <- Map.get(conn.assigns, :auth_tokens, [])
+      |> Enum.map(fn
+        {"bearer", token} -> JWT.parse_token(token)
+        _ -> {:ok, %{}}
+      end)
+      |> Result.list_to_result()
+      |> Result.map_err(fn decode_errors -> {:error, decode_errors} end) do
+
+        # Merge the list of maps into a single map
```
kevinbader: Merge the list of **claims** <-- better to be specific

consider this:
```elixir
auth_tokens = Enum.reduce(auth_tokens, &Map.merge/2)
```

Also, `auth_tokens` is not a good name here, as it implies you're dealing with a list of tokens while in fact this is a merged map of claims.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+  end
+
+  defp decode_metadata(conn) do
+    with {"application", "json"} <- content_type(conn),
+         {:ok, body, conn} <- BodyReader.read_full_body(conn),
+         {:ok, json} <- Jason.decode(body),
+         {:parse, %{"metadata" => metadata}} <- {:parse, json} do
+
+      conf = config()
+      jwt_fields = conf.jwt_fields
```
kevinbader: `jwt_fields` is not a good name for this. It reads like "the fields of a JWT", while in fact this is far from that.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+        end)
+
+        # Get values from JWT
+        jwt_vals = jwt_fields
+        |> Enum.map(fn x ->
+          key = x |> elem(0)
+          jwt_key = x |> elem(1)
+          {key, auth_tokens[jwt_key]}
+        end)
+        |> Enum.into(%{})
```
kevinbader: use `for` instead (this also filters out `nil` values that your code retains):

```elixir
for {key, jwt_key} <- jwt_fields, jwt_val = auth_tokens[jwt_key], into: %{}, do: {key, jwt_val}
```

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+
+        # Merge values from metadata with the values from JWT, prioritizing JWT values
+        metadata = Map.merge(metadata, jwt_vals)
+
+        conn
+        |> Plug.Conn.assign(:metadata, metadata)
+      else
+        {:error, _} ->
+          conn
+          |> Plug.Conn.assign(:metadata, metadata)
```
kevinbader: Extract a function that generates the metadata, given the metadata from the request body plus the list of claims. Write unit tests for this function. Then `assign` the returned `metadata` to `conn` here.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+    else
+      {:parse, json} ->
+        message = """
+        Expected field "metadata" is not present.
+
+        Decoded request body:
+        #{inspect(json)}
+        """
+
+        fail!(conn, message)
```
kevinbader: It's not necessary to `halt` the pipeline here as it's the end of the controller anyway. I'd prefer not to jump around and read the response here directly, so please get rid of `fail!`.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+
+        Technical info:
+        #{inspect(error)}
+        """
+
+        fail!(conn, message)
+    end
+  end
+
+  defp do_set_metadata(%{halted: true} = conn, _connection_id), do: conn
```
kevinbader: will this ever match? If I'm not missing something, a match here would be a logic error. If so, please remove this line.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+        conn
+        |> Plug.Conn.assign(:metadata, metadata)
+      else
+        {:error, _} ->
+          conn
+          |> Plug.Conn.assign(:metadata, metadata)
+      end
+    else
+      {:parse, json} ->
+        message = """
```
kevinbader: The function is called `decode_metadata` and not `decode_metadata_then_assign_it_to_conn_but_if_there_is_an_error_emit_a_bad_request_response`. Consider separating the build-up of `metadata` (see comment above) and handling of `conn`.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+        """
+
+        fail!(conn, message)
+    end
+  end
+
+  defp do_set_metadata(%{halted: true} = conn, _connection_id), do: conn
+
+  defp do_set_metadata(%{assigns: %{metadata: metadata}} = conn, connection_id) do
+    # Check if metadata contains all the indexed fields; if not: send back a bad request
```
kevinbader: again this mixes up the `check` part with the `send` part. Factor out the former into a dedicated function and write a unit test for it, then use the result to figure out whether to send a response to the client.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+    # This needs to be done here because now the values from the JWT got inserted
+    if contains_idx?(metadata) === false do
```
kevinbader: Using `=== false` is extremely uncommon in Elixir code; consider removing it.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+      message = """
+        Metadata doesn't contain indexed fields.
+        """
+
+      fail!(conn, message)
+    else
+      Logger.info("Metadata: " <> inspect(metadata))
```
kevinbader: * Please use `info` level only for important events, not for debugging metadata like this.
* When using `Logger`, _always_ use a function argument. E.g.: `Logger.info(fn -> "foo bar" end)`

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+
+      conf = config()
+
+      indexed_fields =
+        Enum.map(conf.indexed_metadata, fn x ->
+          {x, metadata[x]}
+        end)
```
kevinbader: shorter using `for`:

```elixir
for x <- conf.indexed_metadata, do: {x, metadata[x]}
```

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,178 @@
...
+
+      Logger.info("Indexed fields: " <> inspect(indexed_fields))
+
+      send_resp(conn, :no_content, "")
+    end
+  end
+
+  defp contains_idx?(metadata) do
```
kevinbader: contains an _index_? consider renaming this.

Azer0s:
@kevinbader Currently in the process of adding unit tests. Is there any **good** way too mock the connection?

kevinbader:
> @kevinbader Currently in the process of adding unit tests. Is there any **good** way too mock the connection?

Sure - see `ConnTest`. Anyway, I think you should make sure that the relevant functions don't rely on `conn` directly, so there's no need to mock anything.

kevinbader:
@Azer0s please don't resolve issues yourself but let the reviewer decide whether a raised concern has been resolved or not

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
kevinbader: why/how is this resolved?

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,195 @@
...
+  defp extract_metadata_from_jwt(conn) do
+    with {:ok, claims} <- Map.get(conn.assigns, :auth_tokens, [])
+    |> Enum.map(fn
+      {"bearer", token} -> JWT.parse_token(token)
+      _ -> {:ok, %{}}
+    end)
+    |> Result.list_to_result()
+    |> Result.map_err(fn decode_errors -> {:error, decode_errors} end) do
+      conf = config()
+      metadata_from_jwt = conf.jwt_fields
```
kevinbader: isn't that a mapping between field names in the metadata and field names in the jwt?

Azer0s:
> @Azer0s please don't resolve issues yourself but let the reviewer decide whether a raised concern has been resolved or not

Ok, will do. Sorry about that.

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,195 @@
...
+    Logger.debug(fn -> "Indexed fields: " <> inspect(indexed_fields) end)
+
+    # TODO: Actually set metadata
+  end
+
+  # ---
+
+  defp has_required_indices?(metadata) do
+    conf = config()
+    metadata_keys = Enum.map(metadata, fn x -> x |> elem(0) end)
```
kevinbader: https://hexdocs.pm/elixir/Map.html#keys/1

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,195 @@
...
+    indexed_fields = for x <- conf.indexed_metadata, do: {x, metadata[x]}
+
+    Logger.debug(fn -> "Indexed fields: " <> inspect(indexed_fields) end)
+
+    # TODO: Actually set metadata
+  end
+
+  # ---
+
+  defp has_required_indices?(metadata) do
```
kevinbader: what about `has_required_fields?` instead? Without a descriptions, it's quite hard to figure out what it means for a map to "have indices".

PR #265, opened 2019-09-24
Fix flaky Cypress tests

Azer0s:
## Description

There were some timing problems with Cypress while running in CI which didn't occur while running locally (probably due to the CI instance being slower than our local machines). I fixed this by using the built-in Cypress timeouts.

### DIFF
--- a/examples/tests/cypress/integration/channels.spec.js
+++ b/examples/tests/cypress/integration/channels.spec.js
@@ -6,7 +6,7 @@ describe('Channels', () => {
       cy.subscribe('mike', 'my.public.event');
       // "foo":"bar" due to way how Cypress handles escaping of curly braces
       cy.sendEvent('my.public.event', '"foo":"bar"');
-      cy.assertReceivedEvents('event-log', '(.*my.public.event)(.*"foo":"bar")');
+      cy.assertReceivedEventLog('event-log', '(.*my.public.event)(.*"foo":"bar")');
       cy.disconnect();
     });
 
@@ -16,11 +16,11 @@ describe('Channels', () => {
       cy.subscribe('mike', 'message');
       // "name":"mike","foo":"bar" due to way how Cypress handles escaping of curly braces
       cy.sendEvent('message', '"name":"mike","foo":"bar"');
-      cy.assertReceivedEvents('event-log', 'message', '"name":"mike","foo":"bar"');
+      cy.assertReceivedEventLog('event-log', '"name":"mike","foo":"bar"');
       // send constrained event for different user - John
       cy.sendEvent('message', '"name":"john","foo":"bar"');
       // Mike shouldn't receive any new event
-      cy.wait(2000).assertReceivedEvents(
+      cy.assertReceivedEventLog(
         'event-log',
         '(.*message)(.*"name":"mike","foo":"bar")'
       );
--- a/examples/tests/cypress/integration/channels.spec.js
+++ b/examples/tests/cypress/integration/channels.spec.js
@@ -6,7 +6,7 @@ describe('Channels', () => {
       cy.subscribe('mike', 'my.public.event');
       // "foo":"bar" due to way how Cypress handles escaping of curly braces
       cy.sendEvent('my.public.event', '"foo":"bar"');
-      cy.assertSingleNewEventContainsMessage('event-log', '(.*my.public.event)(.*"foo":"bar")');
+      cy.assertEventLogHasMatch('(.*my.public.event)(.*"foo":"bar")');
       cy.disconnect();
     });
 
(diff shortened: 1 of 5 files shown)


examples/tests/cypress/support/commands.js
```diff
@@ -48,7 +48,15 @@ Cypress.Commands.add('sendEvent', (eventType, message) => {
 
 Cypress.Commands.add('assertReceivedEvents', (element, message) => {
   // assert number and content of received events
-  cy.get(`#${element}`)
+  cy.get(`#${element} > li`, {timeout: 30000})
+    .should('be.visible')
+    .first()
+    .contains(message)
```
kevinbader: the name suggests that it checks a _list_ of _events_ against (a list of) already _received events_. What it _actually_ does is to check whether a given _message_ is _on top_ of the list of messages. Also, the element is not fixed, yet there are some assumptions: it must be a list, and - according to the function name or argument name - it must contain events or messages. Please refactor.

examples/tests/cypress/support/commands.js
```diff
@@ -48,7 +48,15 @@ Cypress.Commands.add('sendEvent', (eventType, message) => {
...
+});
+
+Cypress.Commands.add('assertReceivedEventLog', (element, message) => {
```
kevinbader: I feel that it doesn't make sense to have two functions that are named similarly and almost do the same thing. Please find an API that better fits the has-the-event-been-received use case; passing the `element` is likely not part of it, btw.

examples/tests/cypress/integration/channels.spec.js
```diff
@@ -16,11 +16,11 @@ describe('Channels', () => {
       cy.subscribe('mike', 'message');
       // "name":"mike","foo":"bar" due to way how Cypress handles escaping of curly braces
       cy.sendEvent('message', '"name":"mike","foo":"bar"');
-      cy.assertReceivedEvents('event-log', 'message', '"name":"mike","foo":"bar"');
+      cy.assertReceivedEventLog('event-log', '"name":"mike","foo":"bar"');
       // send constrained event for different user - John
       cy.sendEvent('message', '"name":"john","foo":"bar"');
       // Mike shouldn't receive any new event
-      cy.wait(2000).assertReceivedEvents(
```
mmacai: 2s wait was there to make sure that no events were received even after some time. If you don't wait at least a bit you can't be sure that no events arrived. 2s give us at least some confidence.

Same goes for other occurrences.

examples/tests/cypress/support/commands.js
Azer0s: Maybe something like `assertNewestEventContainsMessage` is better here?

examples/tests/cypress/integration/channels.spec.js
Azer0s: I mean...the timeouts basically achieve the same thing, right? As I see it, `wait` just adds to the tests flakyness.

mmacai: No they don't, timeouts just say how long to wait till e.g. element is visible (e.g. it's not empty but contains one `li` element or whatever). Here we are testing that even after some time there is still one event (so you already start with one `li` and expect it won't be 2 or 3 therefore the wait period). So without waiting, you are actually testing nothing.

Azer0s: The problem is that even 2 seconds don't guarantee anything. If the browser is busy or the CI is just slow, even after 2 seconds of wait time the test isn't sound.

Azer0s: I can put the `wait` back and increase the wait time but we should really be looking for a long term solution.

mmacai: I agree (mentioned it in the first comment), but it's not that easy, especially when cypress doesn't support by default multi-tabs, multi-windows and you have to put it together with Selenium or something.

Azer0s: Should I just re-add the `wait`?

Azer0s: @mmacai I have re-added `wait` for now. The tests are not flaky anymore, which was the priority. I will work on actually fixing the test (with the above mentioned method; two connections) at a later point.

kevinbader:
As discussed out of band, I'd prefer this over the current naming:

```plain
cy.assertEventReceived(<event object>)
cy.assertEventReceivedWithMessage(<text message>)
```

For this to work, I'd expect `event-log` and `eventList` to use the same name (i.e., one of them needs to be renamed).

examples/tests/cypress/integration/channels.spec.js
Azer0s: @mmacai Okay...it got flaky again. Apparently the `wait` does kinda benefit to the test being flaky ☹️

Any ideas on that?

mmacai: Hmm, that's strange, don't see any reason why it would cause flakiness, since it's pretty much just sitting there and doing nothing. Did you save the error log from Travis by any chance?

examples/tests/cypress/support/commands.js
```diff
@@ -46,9 +46,17 @@ Cypress.Commands.add('sendEvent', (eventType, message) => {
...
+Cypress.Commands.add('assertEventInEventList', (message) => {
   // assert number and content of received events
-  cy.get(`#${element}`)
+  cy.get(`#eventList > li`, {timeout: 30000})
+    .should('be.visible')
+    .first()
+    .contains(message)
+});
+
+Cypress.Commands.add('assertEventLogHasMatch', (message) => {
```
mmacai: The channels example is failing, because we have here `#${element}`, but there is no such argument, only `message`. So, we should either pass name of the element from tests or set it to `#event-log`.

PR #266, opened 2019-10-02
154: VConnection, reconnect, connection init

Azer0s:
## Description

To implement the session API, we needed to keep track of connection state, which, in turn, meant rewriting connection handling into a more abstract, generalized solution (`VConnection`).

Creating this VConnection allowed me to do all sorts of other cool stuff (pre-initialization of a connection and reconnects) which now segued into this PR.

As for use cases:

### Use case: Reconnect

Bob has the website of service x open on his laptop while sitting in a train from location a to location b when suddenly the train goes into a tunnel. Bob loses his internet connection. When the train leaves the tunnel, Bobs laptop regains internet connection. Lucky for Bob, service x uses RIG and thus Bob is able to pick up his session where he left off.

### Use case: Pre-initialization of VConnection

This is more of a technical use case that allows for some nice frontend code when dealing with reconnections.

Something like:

```js
const baseUrl = "http://localhost:4000/_rig/v1"
let response = await fetch(`${baseUrl}/connection/init`, {
  method: "GET",
  mode: "cors"
})
const token = await response.text()

const connection_token = response.status == 200 ? `?connection_token=${token}` : ""
const eventUrl = `${baseUrl}/connection/sse${connection_token}`

source = new EventSource(eventUrl)
...
```

## Metadata

Please take a look at #262

## What to look out for

Dear reviewer, I want you to

- help to find bugs


lib/rig_inbound_gateway_web/connection_init.ex
```diff
@@ -47,10 +49,35 @@ defmodule RigInboundGatewayWeb.ConnectionInit do
...
+          # and let VConnection handle subscription refresh and heartbeat
+          VConnection.start(self(), subscriptions, heartbeat_interval_ms, subscription_refresh_interval_ms)
+        {:ok, pid} ->
+          # We're going to try to reconnect
+          if Process.alive?(pid) do
+            # Try a reconnect
+            vpid = GenServer.call(pid, :reconnect)
+
+            # Re-register subscriptions
+            send pid, :set_subscriptions
```
kevinbader: if the client is reconnecting and passes a different set of subscriptions, the newer set should replace the existing subscriptions

lib/rig_inbound_gateway_web/cors.ex
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def cors do
+    quote do
+      defp with_allow_origin(conn) do
+        %{cors: origins} = config()
+        put_resp_header(conn, "access-control-allow-origin", origins)
+      end
+    end
+  end
```
kevinbader: imho this should always be included. in other words, `use Cors` should import that function; only `preflight_put` should be optional

lib/rig_inbound_gateway_web/v1/connection_controller.ex
```diff
@@ -0,0 +1,74 @@
...
+  @doc """
+  ### Dirty Testing
+
+      CONN_TOKEN=$(http :4000/_rig/v1/connection/init)
+      http --stream ":4000/_rig/v1/connection/sse?connection_token=$CONN_TOKEN"
+  """
+  @spec init(conn :: Plug.Conn.t(), params :: map) :: Plug.Conn.t()
+  def init(%{method: "GET"} = conn, _) do
+    {:ok, vconnection_pid} = VConnection.start(@heartbeat_interval_ms, @subscription_refresh_interval_ms)
+    send vconnection_pid, :timeout
```
kevinbader: `:timeout` is also the message that signals a GenServer timeout => please use a different atom for this to avoid confusion

lib/rig_inbound_gateway_web/v1/metadata_controller.ex
```diff
@@ -0,0 +1,220 @@
+defmodule RigInboundGatewayWeb.V1.MetadataController do
```
kevinbader: again, different PR, right?

lib/rig_inbound_gateway_web/v1/websocket.ex
```diff
@@ -84,6 +85,7 @@ defmodule RigInboundGatewayWeb.V1.Websocket do
       {:reply, closing_frame(reason), :no_state}
     end
 
+     # TODO: Handle existing connection token in connection_init; use GenServer.call with a timeout of 500ms to check if VConnection is alive
```
kevinbader: do it

lib/rig_inbound_gateway_web/v1/websocket.ex
```diff
@@ -117,43 +119,33 @@ defmodule RigInboundGatewayWeb.V1.Websocket do
...
+    {:reply, closing_frame("Session killed."), state}
   end
 
   @impl :cowboy_websocket
-  def websocket_info(:refresh_subscriptions, state) do
-    EventFilter.refresh_subscriptions(state.subscriptions, [])
-    Process.send_after(self(), :refresh_subscriptions, @subscription_refresh_interval_ms)
-    {:ok, state}
+  def websocket_info(:close, state) do
+    # TODO: Test
```
kevinbader: do it

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
+defmodule RigInboundGatewayWeb.VConnection do
+  @moduledoc """
+  The VConnection acts as an abstraction for the actual connections.
+
+  It monitors the processes, sends out heartbeats and subscription refreshes and sends out events to the serverless engine when intialized, connection is lost or connection is re-established.
```
kevinbader: serverless engine? that's outside RIG's concern

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
...
+  alias Rig.EventFilter
+  alias Rig.RingBuffer
+  alias RigCloudEvents.CloudEvent
+  alias RigInboundGateway.Events
+
+  require Logger
+
+  use GenServer
+
+  # TODO: Discuss a reasonable kill delay
```
kevinbader: should be 10-15mins imho

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
...
+  @vconnection_kill_delay 120_000
+
+  # TODO: Discuss a reasonable buffer size
```
kevinbader: should probably be configurable at some point but I guess 100 is an okay number for now

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
...
+  alias Rig.RingBuffer
+  alias RigCloudEvents.CloudEvent
+  alias RigInboundGateway.Events
+
+  require Logger
+
+  use GenServer
+
+  # TODO: Discuss a reasonable kill delay
+  @vconnection_kill_delay 120_000
```
kevinbader: please generally append the unit to the variable name, e.g., `_s` to signal seconds and `_ms` to signal milliseconds.

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
...
+
+    if state.monitor do
+      # Demonitor old connection in case we replace an existing connection
+      Process.demonitor(state.monitor)
+    end
+
+    # In case an old connection exists, kill it off before replacing it
+    send! state.target_pid, :close
+
+    {pid, _} = from
```
kevinbader: calling everything `pid` makes the code rather hard to read. what about `{from_pid, _} = from`?

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
...
+    # Replace current subscriptions:
+    state = Map.put(state, :subscriptions, subscriptions)
+
+    send! state.target_pid, {:forward, Events.subscriptions_set(subscriptions)}
+
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info({:set_metadata, metadata}, state) do
```
kevinbader: again not part of this PR

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
...
+  end
+
+  defp send!(pid, data)
+  defp send!(nil, _data), do: nil
+  defp send!(pid, data), do: send pid, data
+
+  defp start_timer(state) do
+    timer = Process.send_after(self(), :kill, @vconnection_kill_delay)
+    state = Map.put(state, :kill_timer, timer)
+    state
```
kevinbader: remove assignment as `Map.put` already returns the state..

lib/rig_inbound_gateway_web/vconnection.ex
```diff
@@ -0,0 +1,204 @@
...
+
+    # TODO: Publish down event
+
+    {:noreply, start_timer(state)}
+  end
+
+  defp create_monitor(state) do
+    monitor = Process.monitor(state.target_pid)
+    state = Map.put(state, :monitor, monitor)
+    state
```
kevinbader: remove assignment as `Map.put` already returns the state..

test/connection_init_test.exs
```diff
@@ -0,0 +1,40 @@
...
+
+  require Logger
+
+  alias HTTPoison
+  alias Rig.Connection.Codec
+
+  @event_hub_http_port Confex.fetch_env!(:rig, RigInboundGatewayWeb.Endpoint)[:http][:port]
+
+  describe "Initialize connection" do
+    test "then connect to it" do
```
kevinbader: the string (test name) should say what is expected. ex_unit doesn't lend itself very well to that given-when-then style of using describe blocks, as you cannot nest test code sufficiently. instead, use describe blocks sparingly and rely on the test's name/description to signal the test's intent. For example:

```elixir
test "An SSE connection can be initialized in advance." do
  ...
end
```

test/connection_init_test.exs
```diff
@@ -0,0 +1,40 @@
...
+      SseClient.disconnect(client)
+    end
+  
+    @tag timeout: 130_000
+    test "then wait for it to time out and connect to it" do
+      url = "http://localhost:#{@event_hub_http_port}/_rig/v1/connection/init"
+      %HTTPoison.Response{body: body} = HTTPoison.get!(url)
+  
+      # Here, we actually want to test the timeout and not destroy the VConnection
+      :timer.sleep(125_000)
```
kevinbader: seriously?

test/connection_test.exs
```diff
@@ -5,20 +5,7 @@ defmodule RigInboundGateway.ConnectionTest do
...
-    dispatch = :cowboy_router.compile(@dispatch)
-    server_name = __MODULE__
-    {:ok, _pid} = :cowboy.start_clear(server_name, [port: @port], %{env: %{dispatch: dispatch}})
-
-    on_exit(fn ->
-      :cowboy.stop_listener(server_name)
-    end)
-
-    :ok
-  end
```
kevinbader: did you remove this by accident? if not, I'd be curious as to why this is no longer needed

test/metadata_controller_test.exs
```diff
@@ -0,0 +1,45 @@
+defmodule RigInboundGatewayWeb.V1.MetadataControllerTest do
```
kevinbader: wrong PR

test/reconnect_test.exs
```diff
@@ -0,0 +1,44 @@
...
+  require Logger
+
+  alias HTTPoison
+  alias Rig.Connection.Codec
+
+  @event_hub_http_port Confex.fetch_env!(:rig, RigInboundGatewayWeb.Endpoint)[:http][:port]
+
+  test "Initialize connection, then disconnect and reconnect" do
+    assert {:ok, client1} = SseClient.connect()
+    {event1, _} = SseClient.read_welcome_event(client1)
```
kevinbader: the idea was to use the returned client for the next interactions... if you do it like this, any future updates to SseClient that make use of this will break this code in subtle ways

test/reconnect_test.exs
```diff
@@ -0,0 +1,44 @@
...
+  use ExUnit.Case, async: true
+
+  require Logger
+
+  alias HTTPoison
+  alias Rig.Connection.Codec
+
+  @event_hub_http_port Confex.fetch_env!(:rig, RigInboundGatewayWeb.Endpoint)[:http][:port]
+
+  test "Initialize connection, then disconnect and reconnect" do
```
kevinbader: related to my other comment on test descriptions, this should describe what the test is actually testing. in this case, it seems to test that "Passing a connection token when connecting via SSE returns the passed connection token.". If that doesn't fit what you had in mind, you should probably think about the test's assertions and update the test description/name accordingly.

test/reconnect_test.exs
```diff
@@ -0,0 +1,44 @@
...
+    assert {:ok, client2} = SseClient.connect([connection_token: event1["data"]["connection_token"]])
+    {event2, _} = SseClient.read_welcome_event(client2)
+    SseClient.disconnect(client2)
+
+    assert Codec.deserialize(event1["data"]["connection_token"]) == Codec.deserialize(event2["data"]["connection_token"])  
+  end
+
+  test "Initialize connection, disconnect, destroy the VConnection and reconnect" do
+    assert {:ok, client1} = SseClient.connect()
+    {event1, _} = SseClient.read_welcome_event(client1)
```
kevinbader: see above

test/reconnect_test.exs
```diff
@@ -0,0 +1,44 @@
...
+    SseClient.flush_mailbox()
+
+    assert {:ok, client2} = SseClient.connect([connection_token: event1["data"]["connection_token"]])
+    {event2, _} = SseClient.read_welcome_event(client2)
+    SseClient.disconnect(client2)
+
+    assert Codec.deserialize(event1["data"]["connection_token"]) == Codec.deserialize(event2["data"]["connection_token"])  
+  end
+
+  test "Initialize connection, disconnect, destroy the VConnection and reconnect" do
```
kevinbader: see above (it actually tests the exact same thing and not - as you've probably intended it to do - that the connection actually matches the previous one. you'll likely need to prove that using an event subscription that is still present etc.)

examples/sse-4-demo-connection-init.html
```diff
@@ -0,0 +1,152 @@
...
+        })
+    }
+
+    function disconnect() {
+      fetch(`${baseUrl}/connection/${connectionToken}/destroy`, {method: "PUT", mode: "cors"}).then(msg => { return msg.text() })
+    }
+
+    function disconnect_connection() {
+      fetch(`${baseUrl}/connection/${connectionToken}/destroy/connection`, {method: "PUT", mode: "cors"}).then(msg => { return msg.text() })
+    }
```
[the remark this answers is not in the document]
Azer0s: Yes. I knew that I was going to rework this. `/destroy` destroys the VConnection and `/destroy/connection` destroys the actual SSE or WS connection.

How about
`DELETE /connection/<token>` to destroy the actual connection
and
`DELETE /connection/<token>/vconnection` to destroy the VConnection?

lib/rig_inbound_gateway_web/v1/connection_controller.ex
Azer0s: How about `:vconnection_timeout`?

lib/rig/connection/codec.ex
```diff
@@ -17,6 +17,8 @@ defmodule Rig.Connection.Codec do
 
   # ---
 
+  def deserialize(nil), do: {:error, nil}
```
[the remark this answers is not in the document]
Azer0s: But if `binary` is nil, this raises an ArgumentError. I'll stick with `{:error, :invalid_term}`.

lib/rig_inbound_gateway_web/vconnection.ex
Azer0s: That long? Isn't that going to use many resources?

lib/rig_inbound_gateway_web/vconnection.ex
Azer0s: Via an environment variable or via config files?

test/connection_init_test.exs
Azer0s: Well...we do need a way to test if the timeout works, don't we? And tbh I haven't figured out a way to programmatically set `@vconnection_kill_delay_ms` to something 0.

examples/sse-4-demo-connection-init.html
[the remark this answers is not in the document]
kevinbader: Let's take a step back - what is the use case for any of those?

lib/rig/connection/codec.ex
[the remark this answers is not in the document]
kevinbader: Please give it a shot anyway. A guard clause that raises an error simply doesn't match; we've used this pattern at numerous places in the code base.

lib/rig_inbound_gateway_web/connection_init.ex
```diff
@@ -40,17 +42,46 @@ defmodule RigInboundGatewayWeb.ConnectionInit do
...
+          if Process.alive?(pid) do
+            # Try a reconnect
+            vpid = GenServer.call(pid, :reconnect)
+
+            if subscriptions != [] do
+              # If the client reconnects with new subscriptions, replace the existing subscriptions
+              send pid, {:set_subscriptions, subscriptions}
+            else
+              # Re-register subscriptions
+              send pid, :set_subscriptions
```
kevinbader: what if I wanted to clear out my subscriptions? as I said, the newer set should always replace the existing subscriptions.

lib/rig_inbound_gateway_web/v1/connection_controller.ex
kevinbader: fine for me

lib/rig_inbound_gateway_web/v1/connection_controller.ex
```diff
@@ -0,0 +1,76 @@
...
+  @doc """
+  ### Dirty Testing
+
+      CONN_TOKEN=$(http :4000/_rig/v1/connection/init)
+      http --stream ":4000/_rig/v1/connection/sse?connection_token=$CONN_TOKEN"
+  """
+  @spec init(conn :: Plug.Conn.t(), params :: map) :: Plug.Conn.t()
+  def init(%{method: "GET"} = conn, _) do
+    {:ok, vconnection_pid} = VConnection.start(@heartbeat_interval_ms, @subscription_refresh_interval_ms)
+    send vconnection_pid, :vconnection_timeout
```
kevinbader: why does the VConnection require its timer to be triggered from outside and doesn't start it automatically?

lib/rig_inbound_gateway_web/vconnection.ex
kevinbader: true. what about 5 mins then? could be a good compromise. given that on a regular connection a user should see something around 5 events every 10 seconds, the event buffer size should be set 150 to reflect this.

lib/rig_inbound_gateway_web/vconnection.ex
kevinbader: via environment variable, similar to how (almost) everything else is configured ;)

test/connection_init_test.exs
kevinbader: see how it's done for the extractor config, for example

examples/sse-4-demo-connection-init.html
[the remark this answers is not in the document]
Azer0s: To gracefully destroy VConnections. So we don't have any open threads after a client is **really** done with whatever they're doing.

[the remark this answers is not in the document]
Azer0s: So far, we don't know if a client just disconnected or really wants to close the session. This was my attempt of resolving that issue.

lib/rig/connection/codec.ex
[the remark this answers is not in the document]
Azer0s: So I should just add `when byte_size(binary) > 0` on the actual function?

lib/rig_inbound_gateway_web/vconnection.ex
Azer0s: 5 mins sounds like a reasonable default. Shouldn't we offer the option to configure this tho?

lib/rig_inbound_gateway_web/vconnection.ex
Azer0s: Ok. I'm on it.

lib/rig_inbound_gateway_web/connection_init.ex
Azer0s: Well...if you wanted to clear out your subscriptions, wouldn't you send an empty subscription  set? If nothing is sent, we just take the old subscriptions. That seems like intended behavior to me. If one **really** wants to clear out their subscriptions, sending an empty subscription set seems like a reasonable way to do so.

test/reconnect_test.exs
Azer0s: What do you mean? 😅

.gitignore
```diff
@@ -39,3 +39,9 @@ erl_crash.dump
 /.elixir_ls/
 *~
 *.orig
+
+*.iml___jb_old___
+
+*.iml
+
+.idea/
```
Azer0s: This is just for people who use IntelliJ to work on RIG (I tried writing Elixir in IntelliJ. It works. It wasn't for me but I thought that if someone does like it, having the IntelliJ project files in our gitignore doesn't harm anyone).

lib/rig/connection/codec.ex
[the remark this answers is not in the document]
kevinbader: yes, but you need to also move the other clause below the main clause, and change it to

```elixir
def deserialize(_), do: {:error, :invalid_term}
```

test/reconnect_test.exs
kevinbader: the returned tuple is `{event, client}`, where `event` is the event read from the stream and `client` is the _potentially updated_ client reference. So you could do `{event1, client1} = SseClient.read_welcome_event(client1)`

lib/rig/connection/codec.ex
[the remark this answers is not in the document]
Azer0s: Ah okay. Didn't know that. Will refactor.

lib/rig_inbound_gateway_web/connection_init.ex
kevinbader: isn't `if subscriptions != []` exactly about the empty subscription set?

Azer0s: You're right. I'd need a way of checking whether `subscriptions` is set.

Azer0s: Ok. Fixed this.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New API for querying and updating the session blacklist: `/v2/session-blacklist`, which introduces the following breaking changes:
   - When a session has been added to the session blacklist successfully, the endpoint now uses the correct HTTP status code "201 Created" instead of "200 Ok".
   - When using the API to blacklist a session, the `validityInSeconds` should now be passed as an integer value (using a string still works though).
+- Added abstraction over SSE and WS (VConnection). This abstraction keeps track of state, manages timers, can be initialized before a client connects to RIG and can be used for reconnects (client can pick up where they left of). This means that for SSE and WS connections, the connection token now refers to the VConnection.
+  - With `GET /v1/connection/init`, one can now pre-initialize this connection
+  - `DELETE /v1/connection/<connection_pid>/vconnection` gracefully destroys this abstraction
+  - `DELETE /v1/connection/<connection_pid>` destroys the actual (WS/SSE) connection
```
kevinbader: Remember that we write the Changelog for our _end users_. To them, the connection is still a connection, and they likely don't care about "abstractions", "VConnection", etc. What is important to them is that RIG now compensates a connection loss by re-sending lost messages. They're interested in the limitations of this.

Of course, the end users are also interested in how to use it, which you address by describing the available endpoints. This is fine in general, but ask the person next to you whether they would know how to use this after reading through it. Also, please remove the endpoint that we've specifically excluded from the API in our discussion yesterday ;)

lib/rig_inbound_gateway_web/router.ex
```diff
@@ -12,30 +12,43 @@ defmodule RigInboundGatewayWeb.Router do
...
-        # up using the Cowboy dispatch configuration; see the `config.exs` file.
+      scope "/connection" do
+        get("/init", ConnectionController, :init)
+
+        scope "/:connection_id" do
+          delete("/", ConnectionController, :destroy_connection)
+          options("/", ConnectionController, :handle_preflight)
+
+          delete("/vconnection", ConnectionController, :destroy)
+          options("/vconnection", ConnectionController, :handle_preflight)
```
kevinbader: please remove this, as discussed

lib/rig_inbound_gateway_web/v1/connection_controller.ex
```diff
@@ -0,0 +1,76 @@
+defmodule RigInboundGatewayWeb.V1.ConnectionController do
+  @moduledoc false
+  use Rig.Config, [:cors]
+  use RigInboundGatewayWeb, :controller
+  use RigInboundGatewayWeb.Cors, :cors
+
+  # TODO: Find a nicer way to do this...maybe `use RigInboundGatewayWeb.Cors, :put, :delete` ?
```
kevinbader: 👍

test/reconnect_test.exs
kevinbader: in other words: the test name describes the steps, not the expectation

lib/rig_inbound_gateway_web/vconnection.ex
kevinbader: yes, good idea. What about `IDLE_CONNECTION_TIMEOUT`?

test/connection_test.exs
Azer0s: This caused **a lot** of problems when running locally. I removed it and it worked. I didn't mean to commit this but ever since, I didn't get any further problems with this test being flaky. When explicitly running this test (`mix test test/connection_test.exs`) I also don't have any problems (when compared to before). But since I don't think that me randomly deleting code is a fix (and I don't really know why this is "fixed"), I will add this back.

lib/rig_inbound_gateway_web/cors.ex
```diff
@@ -1,41 +1,36 @@
...
   end
 
-  def preflight_put do
+  def cors_preflight(which) do
+    methods = which
+    |> Enum.map(fn x ->
+      Atom.to_string(x)
+      |> String.upcase
+    end)
+    |> Enum.join(",")
```
kevinbader: nitpicking, aber imho könnte man das idiomatischer schreiben:

```elixir
def cors_preflight(methods) do
  methods =
    methods
    |> Enum.map(&Atom.to_string/1)
    |> Enum.map(&String.upcase/1)
    |> Enum.join(",")
```

test/connection_init_test.exs
```diff
@@ -0,0 +1,41 @@
...
+  
+      assert {:ok, client} = SseClient.connect([connection_token: body])
+      {event, client} = SseClient.read_welcome_event(client)
+  
+      assert Codec.deserialize(event["data"]["connection_token"]) == Codec.deserialize(body)
+  
+      SseClient.disconnect(client)
+    end
+  
+    @tag timeout: 130_000
```
mmacai: Is this still needed?

lib/rig_inbound_gateway_web/vconnection.ex
mmacai: Can we put back `100` as discussed? Right now there is just `1`.

CHANGELOG.md
mmacai: Agree, can we adjust the text to be more straightforward for users? Also, don't know which endpoint was discussed.

mmacai:
Please run `mix format`, lot of code is not formatted according to built-in Elixir standards/rules.

examples/sse-4-demo-connection-init.html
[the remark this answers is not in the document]
mmacai: @kevinbader does the current form make sense for you or you would do some adjustments?

lib/rig_inbound_gateway_web/vconnection.ex
Azer0s: @mmacai It is `1` because the resend feature is yet to be implemented/discussed. The buffer is, essentially, just stub.

test/connection_init_test.exs
Azer0s: No. I will remove this.

lib/rig_inbound_gateway_web/vconnection.ex
mmacai: Ok.

examples/sse-4-demo-connection-init.html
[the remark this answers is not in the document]
mmacai: Per my information -- resolved internally.

CHANGELOG.md
mmacai: Per my information -- resolved internally.

kevinbader: nope, not resolved

examples/sse-4-demo-connection-init.html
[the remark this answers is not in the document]
kevinbader: nope, not resolved - deleting individual connections shouldn't be part of the external API, as discussed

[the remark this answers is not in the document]
Azer0s: @kevinbader Sorry. Will fix as soon as I get to it.

PR #269, opened 2019-10-31
WIP: Add support for CloudEvents 1.0

### LINKED ISSUE #264: Support CloudEvents v1.0-rc1
The first release candidate of the spec! Even though I would've liked them to use proper semantic versioning ;P

https://github.com/cloudevents/spec/releases/tag/v1.0-rc1

kevinbader:
Closes #264

### DIFF
--- a/lib/rig_cloud_events/cloud_event.ex
+++ b/lib/rig_cloud_events/cloud_event.ex
@@ -70,6 +70,7 @@ defmodule RigCloudEvents.CloudEvent do
 
   def specversion(%__MODULE__{parsed: parsed}) do
     cond do
+      specversion_1_0?(parsed) -> {:ok, "1.0"}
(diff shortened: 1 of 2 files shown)


mmacai:
Can we get a changelog entry?

mmacai:
Tested with our simple sse html demo -- works without any problem. However, I searched for `0.2` (`0.1` has similar search results) to see where we are using ce versions and there are couple of places that might need an update:

- lib/rig_inbound_gateway_web/v1/sse.ex:184
- lib/rig_inbound_gateway/events.ex:33
- lib/rig_inbound_gateway/request_logger/kafka.ex:50
- lib/rig_kafka/client.ex:59
- lib/rig/sources/http/handler.ex:44
- test/support/clients.ex:115

PR #290, opened 2020-03-13
285 - Update Elixir to 1.10 & dependencies

### LINKED ISSUE #285: Update to Elixir 1.10

mmacai:
## Description

- increased Elixir version to 1.10 + did some tiny necessary updates
- updated Elixir libraries
- updated NPM libraries and removed Eslint as it is not really needed and generates vulnerabilities
- updated Node in Dockerfiles to 10 as 8 is no longer supported

Closes #285 

## What to look out for

Dear reviewer, I want you to

Besides that:
- run all type of tests
- check logs - at least Kafka logs should look a bit differently
- in general rig functionality should be unchanged

## Process

Have fun!


examples/channels-example/service/Dockerfile
```diff
@@ -10,7 +10,7 @@ ENV NODE_ENV=production
 COPY package.json /opt/service/package.json
 
 # Install all dependencies
-RUN npm i
+RUN npm i --no-package-lock
```
kevinbader: why?

lib/rig/distributed_set.ex
```diff
@@ -258,11 +258,9 @@ defmodule RIG.DistributedSet do
...
     n_deleted = :ets.select_delete(ets_table, match_spec)
 
-    Logger.debug(fn ->
-      if n_deleted > 0,
-        do: "Removed #{n_deleted} expired records",
-        else: :skip
-    end)
+    if n_deleted > 0 do
+      Logger.debug(fn -> "Removed #{n_deleted} expired records" end)
+    end
```
kevinbader: hm easier to read but runtime performance "hit". Probably neglectable though..

mix.exs
```diff
@@ -90,56 +90,56 @@ defmodule RIG.MixProject do
...
       # Linting:
-      {:credo, "~> 1.0", only: [:dev, :test], runtime: false},
+      {:credo, "~> 1.3", only: [:dev, :test], runtime: false},
       # Static type checks:
       {:dialyxir, "~> 1.0.0-rc.6", only: [:dev], runtime: false},
       # OTP releases:
       {:distillery, "~> 2.1", runtime: false},
       # Documentation generator:
-      {:ex_doc, "~> 0.16", only: :dev, runtime: false},
+      {:ex_doc, "~> 0.21", only: :dev, runtime: false},
```
kevinbader: would it make sense, for `excoveralls`, `credo`, `dialyxir` and `ex_doc` to use `>=` instead of `~>`?

examples/channels-example/service/Dockerfile
mmacai: There were some warnings in logs + it doesn't really make sense to generate lock file in this case imo.

lib/rig/distributed_set.ex
mmacai: This was kind of forced with the new Elixir version, otherwise it was crashing with something like "No clause match for skip". Have better idea how to make it work?

mix.exs
mmacai: Dunno, tried to stick with what we have. What is your opinion?

lib/rig/distributed_set.ex
kevinbader: hm checked the implementation of `Logger` - `:skip` is not supported anymore since 1.10.

mix.exs
kevinbader: I guess it'd make sense cause any newer version of those tools is better than their older versions

mmacai: Ok, I'll check on Friday.

PR #296, opened 2020-03-21
fix website navigation on intro.html

Knappek:
Small fix of the documentation website navigation.


Azer0s:
I think we can add the yarn.lock to gitignore, right?

.gitignore
```diff
@@ -39,3 +39,4 @@ erl_crash.dump
 /.elixir_ls/
 *~
 *.orig
+yarn.lock
```
kevinbader: why?

Knappek: okay, I researched a little whether you should put yarn.lock to gitignore or not. There are different opinions on that, but the [official documentation](https://classic.yarnpkg.com/en/docs/migrating-from-npm) mentions 
> You don’t need to read or understand this file - just check it into source control. When other people start using Yarn instead of npm, the yarn.lock file will ensure that they get precisely the same dependencies as you have.

So let's remove it again from gitignore...

PR #298, opened 2020-04-24
Add JSON log

Azer0s:
## Description

JSON Log is not only a nice to have, for many scenarios it is also an important requirement. The goal of this PR is to introduce two new kinds of loggers into RIG, as well as a new environment variable.

With `LOG_TYPE`, you should be able to control what kind of log RIG produces on STDOUT. This can either be `erlang`(this is the default and is set if the environment variable is not set), `json` (this is basic JSON, nothing too fancy here) and `gcl` (which is the Google Cloud Logger LogEntry format).

## TODO

- [x] GCL
- [x] ~Maybe not use a 3rd party lib (?)~
- [x] Changelog
- [x] Doc

## What to look out for

Dear reviewer, I want you to

## Process

Have fun!


Azer0s:
FYI: GitHub behaves really weird for me. It shows me that I did 9 commits when I only did 1. The other commits are from an entirely different branch. When I try to merge this branch locally, everything behaves as intended and only the 1 commit gets merged in. So this might just be a visual bug.

kevinbader:
> FYI: GitHub behaves really weird for me. It shows me that I did 9 commits when I only did 1. The other commits are from an entirely different branch. When I try to merge this branch locally, everything behaves as intended and only the 1 commit gets merged in. So this might just be a isual bug.

image

It looks like you've based c996dc0a on the vconnection branch. You've merged master into `feature/json-log`, but merging back into master would definitely introduce those additional commits. I suggest cherry-picking c996dc0a onto master.

Azer0s:
> > FYI: GitHub behaves really weird for me. It shows me that I did 9 commits when I only did 1. The other commits are from an entirely different branch. When I try to merge this branch locally, everything behaves as intended and only the 1 commit gets merged in. So this might just be a isual bug.
> 
> image
> 
> It looks like you've based [c996dc0](https://github.com/Accenture/reactive-interaction-gateway/commit/c996dc0a25e8767df775390718273ebd2f09f9d6) on the vconnection branch. You've merged master into `feature/json-log`, but merging back into master would definitely introduce those additional commits. I suggest cherry-picking [c996dc0](https://github.com/Accenture/reactive-interaction-gateway/commit/c996dc0a25e8767df775390718273ebd2f09f9d6) onto master.

Thanks for the tip! This (as you can see 😁) worked!

lib/rig/application.ex
```diff
@@ -2,17 +2,32 @@ defmodule Rig.Application do
...
+      :json ->
+        Logger.add_backend(LoggerJSON)
+      :gcl ->
+        Logger.add_backend(LoggerJSON)
+        Logger.configure_backend(LoggerJSON, formatter: LoggerJSON.Formatters.GoogleCloudLogger)
+      _ ->
+        # TODO: Fix format; context: config.exs:94
+        # Or can I leave it like that? The only difference seems to
+        # be that the level and the module are switched
+        Logger.add_backend(:console)
```
kevinbader, line 28: So the config for console is not used here? You could try to set the values using Application.put_env, which should basically do the same..

It would be good to respect the config setting if possible, because it deduplicates the meta data.

mix.exs
```diff
@@ -53,7 +53,7 @@ defmodule RIG.MixProject do
   def application do
     [
       mod: {Rig.Application, []},
-      extra_applications: [:logger, :runtime_tools, :prometheus_ex, :prometheus_plugs],
+      extra_applications: [:runtime_tools, :prometheus_ex, :prometheus_plugs, :jason, :logger_json],
```
kevinbader, line 56: is this really necessary? especially for `jason`, which we are already using in the code base without including it here

Azer0s: I removed both now. Still works. I had them in because documentation requested it. But if it works, it works I guess 😅

lib/rig/application.ex
Azer0s: What do you mean? I meant that the format specified in config.exs doesn't apply here. I can fix metadata just by setting `:metadata` to `:all`. I don't think I can set the format through an env var.

mix.exs
Azer0s: So...done :)

lib/rig/application.ex
kevinbader: I wasn't talking about environment variables - see https://hexdocs.pm/elixir/Application.html#put_env/4

kevinbader: > I meant that the format specified in config.exs doesn't apply here. I can fix metadata just by setting :metadata to :all.

Why can't you apply the previous config here? Not sure what you mean..

Azer0s: Hmm...it appears to work now...for some reason...

Azer0s: Like...it produces the same result (log-wise) as the master branch

Azer0s:
@kevinbader Shall I maybe add a paragraph in `metrics-details.md`?

lib/rig/application.ex
Azer0s: @kevinbader Is this closable then?

kevinbader: please double check whether it _actually_ works now ;) but yes, if it _really_ works, sure

kevinbader:
@Azer0s please merge with current master.

> @kevinbader Shall I maybe add a paragraph in `metrics-details.md`?

`metrics-details.md` is about Prometheus metrics - what would you like to document there?

kevinbader (approved):

kevinbader (approved):

PR #318, opened 2020-07-22
make it possible to return cached response 

### LINKED ISSUE #300: Allow an asynchronous service to respond immediately
When in the reverse-proxy configuration a route has target `http` and the `response_from` option is present, the synchronous response is dropped.

```plain
+----------+                    +-------+                       +----------------+
|          +------------------->+       +---------------------->+                |
|  Client  |                    |  RIG  |                       |  Microservice  |
|          +<-------------------+       |  X <------------------+                |
+----------+                    +---+---+                       +-------+--------+
               async response       ^      response                     |
                                    |      is dropped                   |
                                    |                                   |
                                    |                                   +
                                    |
                                    |                               later...
                                    |
                                    |                                   +
                                    |                                   |
                                    |         async response            |
                                    |                                   |
                                    +-----------------------------------+

```

The idea is to make this more flexible:

- Simply return the response as if `response_from` was not set
- If the response's status code is 202 Accepted, ignore the response and wait for the result

This way, a microservice can return a cached response or an error immediately:

```plain
+----------+                    +-------+                       +----------------+
|          +------------------->+       +---------------------->+                |
|  Client  |                    |  RIG  |                       |  Microservice  |
|          +<-------------------+       +<----------------------+                |
+----------+                    +-------+                       +----------------+
                                               bad request!
```

We _could_ add this as an option to the current configuration but I think that would make an already complex-looking configuration even weirder. So I propose this as a _breaking change_ to be included in the upcoming 3.0 release.

#### Implementation notes

Take a look at `lib/rig_inbound_gateway/api_proxy/handler/http.ex`

Knappek:
## Description

closes #300 

## What to look out for

Dear reviewer, I want you to

## Process

Have fun!


test/rig_tests/proxy/response_from/async_http_test.exs
```diff
@@ -98,4 +98,145 @@ defmodule RigTests.Proxy.ResponseFrom.AsyncHttpTest do
     # ...but the client got the response sent to the HTTP internal endpoint:
     assert Jason.decode!(res_body)["data"] == async_response
   end
+
+  test_with_server "Given response_from is set to http_async and the backend responds with 200, the http response is synchronous." do
```
kevinbader, line 102: In a similar vein, I would change this to

```elixir
  test_with_server "Given response_from=http_async, when the backend responds with 200, RIG forwards this 200-response (and does not expect an asynchronous response for this request)." do
```

test/rig_tests/proxy/response_from/async_http_test.exs
```diff
@@ -98,4 +98,145 @@ defmodule RigTests.Proxy.ResponseFrom.AsyncHttpTest do
...
+    assert FakeServer.hits() == 1
+    # ...the connection is closed and the status is OK:
+    assert res_status == 200
+    # ...the client receives the synchronous http response:
+    assert Jason.decode!(res_body) == sync_response
+    # ... and never receives the asynchronous response
+    assert Jason.decode!(res_body) != async_response
+  end
+
+  test_with_server "Given response_from is set to http_async and the backend responds with 400, the client should not receive any data." do
```
kevinbader, line 174: While the implementation is correct, this doesn't sound right - the client indeed receives the 400-response, so it receives data; it just happens that the response has no body. I suggest adding a body like "Bad request from the test endpoint.", which would make the test more readable and also make sure that the response the test sees is actually the response we expect, and not some internal error response we get from the framework.

Keeping with the style above, this could read:

```elixir
  test_with_server "Given response_from=http_async, when the backend responds with 400, RIG forwards this 400-response (and does not expect an asynchronous response for this request)." do
```

test/rig_tests/proxy/response_from/async_http_test.exs
```diff
@@ -27,7 +27,7 @@ defmodule RigTests.Proxy.ResponseFrom.AsyncHttpTest do
     end)
   end
 
-  test_with_server "Given response_from is set to http_async, the http response is taken from the internal HTTP endpoint." do
+  test_with_server "Given response_from is set to http_async and the backend responds with 202, the http response is taken from the internal HTTP endpoint." do
```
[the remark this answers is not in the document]
Knappek: agree, I like verbose test descriptions 👍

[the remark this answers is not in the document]
Knappek: the test name was too long, there is a limit in mix, that's why I haven't copied it 1:1

test/rig_tests/proxy/response_from/async_http_test.exs
Knappek: done

test/rig_tests/proxy/response_from/async_http_test.exs
Knappek: done

test/rig_tests/proxy/response_from/async_http_test.exs
```diff
@@ -27,7 +27,10 @@ defmodule RigTests.Proxy.ResponseFrom.AsyncHttpTest do
     end)
   end
 
-  test_with_server "Given response_from is set to http_async, the http response is taken from the internal HTTP endpoint." do
+  # test_with_server "" do
```
kevinbader, line 30: remove this comment?

Knappek: done

Knappek:
> Also, please wait for #321 to land and then update `docs/api-gateway.md` to reflect the changes here. For example:
> 
> > When set to `kafka`, the response to the request is not taken from the `target` (e.g., for `target` = `http` this means the backend's HTTP response is ignored)
> 
> ...which is not correct as the response is not (always) ignored.

I have updated it. In addition to that, I will soon have it documented separately (#330) and will also link it to refer to further details (with a diagram etc.).

docs/api-gateway.md
```diff
@@ -180,7 +180,7 @@ The endpoint expects the following request format:
...
+A behavior like this can be configured using an endpoints' `response_from` property. When set to `kafka`, the response to the request is not necessarily taken from the `target`, e.g., for `target` = `http` this means the backend's HTTP response might be ignored - it's the responsibility of the backend service where to read the response from: If the backend returns with the HTTP code `202 Accepted`, the response will be read from a Kafka topic. If the backend returns a different HTTP code (can be `200` or `400` or whatever makes sense), only then the response will be read synchronously from the http target directly (this allows the backend to return cached responses). In order to enable RIG to correlate the response from the kafka topic with the original request, RIG adds a correlation ID to the request (using a query parameter in case of `target` = `http`, or backed into the produced CloudEvent otherwise). Backend services that work with the request need to include that correlation ID in their response; otherwise, RIG won't be able to forward it to the client (and times out).
```
kevinbader, line 183:
```suggestion
A behavior like this can be configured using an endpoints' `response_from` property. When set to `kafka`, the response to the request is not necessarily taken from the `target`, e.g., for `target` = `http` this means the backend's HTTP response might be ignored - it's the responsibility of the backend service where to read the response from: If the backend returns with the HTTP code `202 Accepted`, the response will be read from a Kafka topic. If the backend returns a different HTTP code (can be `200` or `400` or whatever makes sense), only then the response will be read synchronously from the http target directly (this allows the backend to return cached responses). In order to enable RIG to correlate the response from the kafka topic with the original request, RIG adds a correlation ID to the request (using a query parameter in case of `target` = `http`, or baked into the produced CloudEvent otherwise). Backend services that work with the request need to include that correlation ID in their response; otherwise, RIG won't be able to forward it to the client (and times out).
```

test/rig_tests/proxy/response_from/async_http_test.exs
```diff
@@ -153,4 +155,146 @@ defmodule RigTests.Proxy.ResponseFrom.AsyncHttpTest do
...
+
+    # Now we can assert that...
+    # ...the fake backend service has been called:
+    assert FakeServer.hits() == 1
+    # ...the connection is closed and the status is OK:
+    assert res_status == 200
+    # ...the client receives the synchronous http response:
+    assert Jason.decode!(res_body) == sync_response
+    # ... and never receives the asynchronous response
+    assert Jason.decode!(res_body) != async_response
```
kevinbader, line 228: This line basically says sync_response != async_response (since res_body doesn't change). I think it could be removed

kevinbader (commented):
A typo and one line to remove, other than that LGTM

kevinbader (approved):

Issue #221, opened 2019-06-02
Support HTTP response code with the API's /responses endpoint


kevinbader:
The `/responses` endpoint, which can be used to asynchronously submit responses to incoming HTTP requests, currently doesn't support HTTP response codes. In other words, it's currently not possible to answer such a request with "401 Not authorized", for example.

The reason for this is that we mistakingly expect responses to be sent as CloudEvents (this is also the case for reading responses from Kafka!). Indeed, a response is a _message_ rather than an _event_ and there's no reason at all to expect it to conform to the CloudEvents standard. In fact, RIG doesn't interpret it in any way, so there's no need to put any constraints on the payload format.

To address this I propose the following changes.

# POST API: /v1/responses

Two modes are supported: binary (preferred) and structured.

Since the endpoint hasn't been included in any release so far, we should be able to make the change without caring about backwards compatibility. We should note down how to migrate any early-adopter code in the Changelog, however.

## Binary Mode

The HTTP header contains prefixed meta data:

```plaintext
rig-correlation: the correlation ID as added by RIG to the forwarded request
rig-response-code: 200
```

The body contains the payload in the encoding specified by the "content-type" header:

```plaintext
in binary format; RIG forwards this as-is
```

## Structured Mode

The header is not used, except for the content-type, which must be "application/json".

Body:

```plaintext
{
  "correlation": "the correlation ID as added by RIG to the forwarded request",
  "response-code": 200,
  "content-type": "text/plain",
  "content": "in whatever encoding is specified in content-type + base64; RIG base64-decodes this string and forward this as the response body with the content-type set to what's specified in this body"
}
```

# Reading the response from Kafka topic

Should work the same way. ~~Kafka headers become a requirement.~~ With support for both the binary and the structured transport modes, using Kafka headers is preferred but not required.

To handle the breaking change, we have two options:

1. Deprecate the old way and use a new topic for responses that follow the new format. Remove support for the deprecated topic/format with 3.0.
1. Also deprecate the old way, but support it alongside the new format using the same topic. The code could distinguish the request formats by looking for `rig-correlation` in the header.

# Reading the response from Kinesis

Kinesis doesn't support headers, so Kinesis only supports the structured transport mode as described above.

## Why is binary mode preferred?

Cause structured mode requires binary payloads to be encoded (e.g. using base 64) in order for it to be sent as a string that is valid in JSON terms. This takes up CPU and makes messages larger.

# References

[...]

PR #321, opened 2020-07-30
221 response codes

mmacai:
## Description

Support custom response code for these `response_from` targets:

- [x] Kafka
- [x] HTTP async
- [-] ~~NATS~~ - NATS -> NATS not implemented as I wasn't able to test it properly
- [-] ~~Kinesis~~ - Not supported atm

## What to look out for

- tests should cover majority of cases and docs should reflect the proper usage/state

Closes https://github.com/Accenture/reactive-interaction-gateway/issues/221

### DIFF
--- a/docs/api-gateway.md
+++ b/docs/api-gateway.md
@@ -212,7 +212,7 @@ Configuration of such API endpoint might look like this:
 - Kinesis -> **not supported**
 - Nats -> `nats`
 
-`http_async` means that correlated response has to be sent to internal `:4010/v2/responses` `POST` endpoint.
+`http_async` means that correlated response has to be sent to internal `:4010/v3/responses` `POST` endpoint.
 
 #### Supported formats
 
@@ -223,9 +223,10 @@ Message headers:
 ```plaintext
 rig-correlation: "correlation_id_sent_by_rig"
 rig-response-code: "201"
+content-type: "application/json"
 ```
 
-> Both headers are required.
+> All headers are required.
 
 Message body:
 
--- a/docs/api-gateway.md
+++ b/docs/api-gateway.md
@@ -218,9 +218,7 @@ Configuration of such API endpoint might look like this:
 
 #### Supported formats
 
-All `response_from` options are using message structures as below.
-
-##### Binary
+All `response_from` options support only binary mode.
 
 Message headers:
 
@@ -229,7 +227,7 @@ rig-correlation: "correlation_id_sent_by_rig"
 rig-response-code: "201"
 ```
 
-> `rig-correlation` is required.
+> Both headers are required.
 
 Message body:
 
@@ -238,27 +236,6 @@ Message body:
   "foo": "bar"
 }
 ```
-
-##### Structured
-
-Message body:
-
-```json
-{
-  "rig": {
-    "correlation": "correlation_id_sent_by_rig",
-    "response_code": 201
-  },
-  "headers": {
-    "foo": "bar"
-  },
-  "body": {
-    "foo": "bar"
-  }
-}
-```
-
-> `rig.correlation` is required.
 
 ## Auth
 
--- a/docs/api-gateway.md
+++ b/docs/api-gateway.md
@@ -214,8 +214,6 @@ Configuration of such API endpoint might look like this:
 
 `http_async` means that correlated response has to be sent to internal `:4010/v2/responses` `POST` endpoint.
 
-`response_from="kafka"` will try to decode Avro encoded message.
-
 #### Supported formats
 
 All `response_from` options support only binary mode.
--- a/docs/api-gateway.md
+++ b/docs/api-gateway.md
@@ -180,9 +180,7 @@ The endpoint expects the following request format:
 ### Wait for response
 
 Sometimes it makes sense to provide a simple request-response API to something that runs asynchronously on the backend. For example, let's say there's a ticket reservation process that takes 10 seconds in total and involves three different services that communicate via message passing. For an external client, it may be simpler to wait 10 seconds for the response instead of polling for a response every other second.
-A behavior like this can be configured using an endpoints' `response_from` property. When set to `kafka`, the response to the request is not taken from the `target` (e.g., for `target` = `http` this means the backend's HTTP response is ignored), but instead it's read from a Kafka topic. In order to enable RIG to correlate the response from the topic with the original request, RIG adds a correlation ID to the request (using a query parameter in case of `target` = `http`, or backed into the produced CloudEvent otherwise). Backend services that work with the request need to include that correlation ID in their response; otherwise, RIG won't be able to forward it to the client (and times out).
-
-> In case you want to use _binary_ transport mode, make sure that `rig` extension (containing correlation ID) is prefixed with `ce_` as well.
+A behavior like this can be configured using an endpoints' `response_from` property. When set to `kafka`, the response to the request is not taken from the `target` (e.g., for `target` = `http` this means the backend's HTTP response is ignored), but instead it's read from a Kafka topic. In order to enable RIG to correlate the response from the topic with the original request, RIG adds a correlation ID to the request (using a query parameter in case of `target` = `http`, or backed into the produced CloudEvent otherwise). **Backend services that work with the request need to include that correlation ID in their response; otherwise, RIG won't be able to forward it to the client (and times out).**
 
 Configuration of such API endpoint might look like this:
 
@@ -207,25 +205,60 @@ Configuration of such API endpoint might look like this:
 
 > Note the presence of `response_from` field. This tells RIG to wait for different event with the same correlation ID.
 
-As an alternative you can set `response_from` to `http_async`. This means that correlated response has to be sent to internal `:4010/v2/responses` `POST` endpoint with a body like this:
+#### Supported combinations (`target` -> `response_from`)
+
+- HTTP -> `kafka`/`http_async`/`kinesis`
+- Kafka -> `kafka`
+- Kinesis -> **not supported**
+- Nats -> `nats`
+
+`http_async` means that correlated response has to be sent to internal `:4010/v2/responses` `POST` endpoint.
+
+`response_from="kafka"` will try to decode Avro encoded message.
+
+#### Supported formats
+
+All `response_from` options are using message structures as below.
+
+##### Binary
+
+Message headers:
+
+```plaintext
+rig-correlation: "correlation_id_sent_by_rig"
+rig-response-code: "201"
+```
+
+> `rig-correlation` is required.
+
+Message body:
 
 ```json
 {
-  "id": "1",
-  "specversion": "0.2",
-  "source": "my-service",
-  "type": "com.example",
-  "rig": {
-    "correlation": "_id_"
-  },
-  "data": {
-    ...
-  }
-  ...
+  "foo": "bar"
 }
 ```
 
-> **NOTE:** Kinesis doesn't support `response_from` field yet.
+##### Structured
+
+Message body:
+
+```json
+{
+  "rig": {
+    "correlation": "correlation_id_sent_by_rig",
+    "response_code": 201
+  },
+  "headers": {
+    "foo": "bar"
+  },
+  "body": {
+    "foo": "bar"
+  }
+}
+```
+
+> `rig.correlation` is required.
 
 ## Auth
 
--- a/lib/rig_api/v1/responses.ex
+++ b/lib/rig_api/v1/responses.ex
@@ -5,7 +5,8 @@ defmodule RigApi.V1.Responses do
   use RigApi, :controller
   use PhoenixSwagger
 
-  alias RigInboundGateway.ApiProxy.ResponseFromParser
+  alias Rig.Connection.Codec
+  alias RigCloudEvents.CloudEvent
 
   @prefix "/v1"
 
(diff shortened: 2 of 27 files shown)


mmacai:
@kevinbader still have to do some testing and probably missed some places to update, but you can already give it a shot.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
```diff
@@ -0,0 +1,103 @@
...
+  def parse(
+        _headers,
+        %{
+          "body" => body,
+          "rig" => rig_metadata
+        } = message
+      ) do
+    with {:ok, correlation_id} <- Map.fetch(rig_metadata, "correlation"),
+         {:ok, deserialized_pid} <- Codec.deserialize(correlation_id),
+         raw_response_code <- Map.get(rig_metadata, "response_code", @default_response_code),
```
kevinbader, line 30: Not sure a default makes sense in this case. You have to provide that "rig" object anyway because of the correlation ID. Making the response code required at least prevents you from accidentally sending errors as OK - imagine you've accidentally misspelled "response_code" and didn't notice, because RIG just uses 200 anyway.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
```diff
@@ -0,0 +1,103 @@
...
+        headers,
+        message
+      ) do
+    with headers_map <- Enum.into(headers, %{}),
+         {:ok, correlation_id} <- Map.fetch(headers_map, "rig-correlation"),
+         {:ok, deserialized_pid} <- Codec.deserialize(correlation_id),
+         raw_response_code <- Map.get(headers_map, "rig-response-code", @default_response_code),
+         # convert status code to int if needed, HTTP headers can't have number as a value
+         {response_code, _} <- to_int(raw_response_code),
+         response_body <- try_encode(message) do
```
kevinbader, line 58: wouldn't it make more sense to not even touch the body in binary mode? I mean, best would be to forward it verbatim - no decoding or encoding

lib/rig_inbound_gateway/api_proxy/handler/kafka.ex
```diff
@@ -279,7 +270,10 @@ defmodule RigInboundGateway.ApiProxy.Handler.Kafka do
         conn
         |> Tracing.Plug.put_resp_header(Tracing.context())
         |> Conn.put_resp_content_type("application/json")
-        |> Conn.send_resp(:ok, response)
+        |> Map.update!(:resp_headers, fn existing_headers ->
+          existing_headers ++ Map.to_list(response_headers)
+        end)
```
kevinbader, line 275: Doesn't `existing_headers` likely already include a `Content-Type` header? (same comment in http.ex)

lib/rig_inbound_gateway/api_proxy/handler/http.ex
```diff
@@ -116,7 +116,10 @@ defmodule RigInboundGateway.ApiProxy.Handler.Http do
         |> with_cors()
         |> Tracing.Plug.put_resp_header(Tracing.context())
         |> Conn.put_resp_content_type("application/json")
-        |> Conn.send_resp(:ok, response)
+        |> Map.update!(:resp_headers, fn existing_headers ->
+          existing_headers ++ Map.to_list(response_headers)
+        end)
```
kevinbader, line 121: Doesn't existing_headers likely already include a Content-Type header?

I've looked into the HTTP grammar (1.0 and 1.1) and couldn't find anything on how often particular header keys may occur and how to deal with multiple occurrences. Therefore, I assume multiple content-type headers lead to unexpected behavior.

docs/api-gateway.md
```diff
@@ -207,25 +205,60 @@ Configuration of such API endpoint might look like this:
...
+    "correlation": "correlation_id_sent_by_rig",
+    "response_code": 201
   },
-  "data": {
-    ...
+  "headers": {
+    "foo": "bar"
+  },
+  "body": {
+    "foo": "bar"
```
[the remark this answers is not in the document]
mmacai: We decided to drop structured mode, so I'll close this one.

docs/api-gateway.md
```diff
@@ -207,25 +205,60 @@ Configuration of such API endpoint might look like this:
...
+
+All `response_from` options are using message structures as below.
+
+##### Binary
+
+Message headers:
+
+```plaintext
+rig-correlation: "correlation_id_sent_by_rig"
+rig-response-code: "201"
```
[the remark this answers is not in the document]
mmacai: Not at the moment, I would add it once there is some feature request for it.

lib/rig_inbound_gateway/api_proxy/handler/http.ex
mmacai: This was originally supported by structured mode, but as we are dropping it, I'll close this one.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
```diff
@@ -0,0 +1,103 @@
...
+  @spec parse([{String.t(), String.t()}, ...], map()) ::
+          {pid, pos_integer(), any(), map()} | String.t()
+  # structured
+  def parse(
+        _headers,
+        %{
+          "body" => body,
+          "rig" => rig_metadata
+        } = message
+      ) do
```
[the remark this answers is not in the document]
mmacai: Structured will be dropped, so closing this one.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
mmacai: Okay, wanted to make users life easier.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
mmacai: So, I've tried and it worked fine with Kafka, but not with Phoenix as it parsed it regardless of format I've tried.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
```diff
@@ -0,0 +1,103 @@
...
+    with {:ok, correlation_id} <- Map.fetch(rig_metadata, "correlation"),
+         {:ok, deserialized_pid} <- Codec.deserialize(correlation_id),
+         raw_response_code <- Map.get(rig_metadata, "response_code", @default_response_code),
+         # convert status code to int if needed, HTTP headers can't have number as a value
+         {response_code, _} <- to_int(raw_response_code),
```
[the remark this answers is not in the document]
mmacai: Should work now.

docs/api-gateway.md
```diff
@@ -207,26 +205,36 @@ Configuration of such API endpoint might look like this:
...
+- HTTP -> `kafka`/`http_async`/`kinesis`
+- Kafka -> `kafka`
+- Kinesis -> **not supported**
+- Nats -> `nats`
+
+`http_async` means that correlated response has to be sent to internal `:4010/v2/responses` `POST` endpoint.
+
+#### Supported formats
+
+All `response_from` options support only binary mode.
```
kevinbader, line 219: Not an issue, just a thought: is "binary mode" a common term outside the cloudevents spec?

docs/api-gateway.md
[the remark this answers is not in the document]
kevinbader: fine for me

lib/rig_api/v1/responses.ex
```diff
@@ -36,18 +35,16 @@ defmodule RigApi.V1.Responses do
...
+  def create(%{req_headers: req_headers} = conn, message) do
+    case ResponseFromParser.parse(req_headers, message) do
+      {deserialized_pid, response_code, response} ->
+        Logger.debug(fn ->
+          "HTTP response via internal HTTP to #{inspect(deserialized_pid)}: #{inspect(message)}"
+        end)
+
+        send(deserialized_pid, {:response_received, response, response_code})
+        send_resp(conn, :accepted, "message sent to correlated reverse proxy request")
+
```
kevinbader, line 47: I didn't notice that before, but I guess this is not how versions are supposed to work :D I'd not touch v1 and v2 and instead create a v3, since this is clearly a breaking change. With 3.0 we probably no longer care about v1 and v2, so you can remove them in this case, but I'd still call the new behavior v3

lib/rig_inbound_gateway/api_proxy/handler/http.ex
```diff
@@ -116,7 +116,7 @@ defmodule RigInboundGateway.ApiProxy.Handler.Http do
         |> with_cors()
         |> Tracing.Plug.put_resp_header(Tracing.context())
         |> Conn.put_resp_content_type("application/json")
-        |> Conn.send_resp(:ok, response)
+        |> Conn.send_resp(response_code, response)
```
kevinbader, line 119: what about the content type? Ideally, we'd forward the body as-is and also retain the content-type, then the backend can send whatever

lib/rig_inbound_gateway/api_proxy/handler/kafka.ex
```diff
@@ -279,7 +269,7 @@ defmodule RigInboundGateway.ApiProxy.Handler.Kafka do
         conn
         |> Tracing.Plug.put_resp_header(Tracing.context())
         |> Conn.put_resp_content_type("application/json")
-        |> Conn.send_resp(:ok, response)
+        |> Conn.send_resp(response_code, response)
```
kevinbader, line 272: same here

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
kevinbader: we had the same issue with the inbound gateway, where we had to disable Phoenix' parsing of the body entirely. So far we didn't need to do that on rig_api but perhaps it'd make sense for this use case... since I know from experience that this breaks all the other endpoints and is quite a nasty thing to do, I suggest you keep it like this for now.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
```diff
@@ -0,0 +1,55 @@
...
+
+      _ ->
+        message
+    end
+  end
+
+  # ---
+
+  defp to_int(value) when is_integer(value), do: value
+  defp to_int(value) when is_binary(value), do: String.to_integer(value)
```
kevinbader, line 54: may I propose the following:

```elixir
  def to_int(string) when is_binary(string) do
    String.to_integer(string)
  rescue
    ArgumentError -> {:error, {:not_an_integer, string}}
  end
```

1. this doesn't die with the not-helpful ArgumentError
1. providing for `is_integer` is not necessary here and would be an unexpected thing to happen

test/rig_tests/proxy/response_from/kafka_test.exs
```diff
@@ -205,93 +114,109 @@ defmodule RigTests.Proxy.ResponseFrom.KafkaTest do
...
-    # ...the fake backend service has been called:
-    assert FakeServer.hits() == 1
-    # ...the connection is closed and the status is OK:
-    assert res_status == 200
-    # ...but the client got the response sent to the Kafka topic:
-    assert Jason.decode!(res_body) == async_response
+    RigKafka.Client.stop_supervised(kafka_client)
   end
+
+  # @tag :kafka
```
kevinbader, line 124: not sure if on purpose or by accident

docs/api-gateway.md
mmacai: No idea tbh, but I guess it makes sense to be consistent with such terms.

lib/rig_api/v1/responses.ex
mmacai: Okay, but I would keep v1 and v2 and remove them in separate issue as there is more deprecated things to remove.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
mmacai: Okay, I'll give it a try.

test/rig_tests/proxy/response_from/kafka_test.exs
mmacai: On purpose, it's the one where I have issues with Kafka consume clash.

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
```diff
@@ -0,0 +1,64 @@
...
+  # ---
+
+  defp to_int(value) when is_integer(value), do: value
+
+  defp to_int(string) when is_binary(string) do
+    String.to_integer(string)
+  rescue
+    ArgumentError -> {:error, {:not_an_integer, string}}
+  end
+end
```
kevinbader, line 64:
```suggestion
  defp to_int(value) when is_integer(value), do: {:ok, value}

  defp to_int(string) when is_binary(string) do
    {:ok, String.to_integer(string)}
  rescue
    ArgumentError -> {:error, {:not_an_integer, string}}
  end
end
```

lib/rig_inbound_gateway/api_proxy/response_from_parser.ex
```diff
@@ -0,0 +1,64 @@
...
+        message
+      ) do
+    with headers_map <- Enum.into(headers, %{}),
+         {:ok, correlation_id} <- Map.fetch(headers_map, "rig-correlation"),
+         {:ok, deserialized_pid} <- Codec.deserialize(correlation_id),
+         {:ok, raw_response_code} <- Map.fetch(headers_map, "rig-response-code"),
+         # forward extra haeders such as content-type
+         extra_headers <- Map.drop(headers_map, ["rig-correlation", "rig-response-code"]),
+         # convert status code to int if needed, HTTP headers can't have number as a value
+         response_code <- to_int(raw_response_code),
```
kevinbader, line 27:
```suggestion
         {:ok, response_code} <- to_int(raw_response_code),
```

PR #327, opened 2020-09-09
make github pages url Accenture upper-case as helm can't handle it

Knappek:
## Description

This does not work:

```shell
helm repo add accenture https://accenture.github.io/reactive-interaction-gateway
```

but this works:

```shell
helm repo add accenture https://Accenture.github.io/reactive-interaction-gateway
```

:man_facepalming:


kevinbader (approved):
lol

PR #329, opened 2020-09-09
Fix restore helm index yml

Knappek:
## Description

Fixes #328 which failed, see [this travis output](https://travis-ci.org/github/Accenture/reactive-interaction-gateway/jobs/725564918).


.travis.yml
```diff
@@ -55,7 +57,20 @@ jobs:
...
+      language: minimal
+      script:
+        - git config --global user.name "${GH_NAME}"
+        - git config --global user.email "${GH_EMAIL}"
+        - echo "machine github.com login ${GH_NAME} password ${GH_TOKEN}" > ~/.netrc
+        - git fetch --unshallow
+        - git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
+        - git fetch origin
+        - git checkout gh-pages
+        - git checkout $(git rev-list -n1 HEAD -- index.yaml)~1 -- index.yaml > /dev/null 2>&1 | true
```
kevinbader, line 70: why ignore stderr here?

kevinbader (commented):
other than the one comment it looks fine

.travis.yml
Knappek: as I wanted this stage to also work if the HEAD already has index.yaml and `$(git rev-list -n1 HEAD -- index.yaml)~1` doesn't. It won't happen actually, but it happened during testing. Hmm, actually this was the explanation of `| true`. We could remove ignoring stderr, but then we could see an error but the pipeline doesn't fail.

kevinbader: ah sure, the file might exist... perhaps a more explicit way of handling this case would be:

```bash
test -e index.yaml || git checkout $(git rev-list -n1 HEAD -- index.yaml)~1 -- index.yaml
```

Knappek: I like that... will change it.

kevinbader (approved):

Issue #278, opened 2019-12-13
Drop support for deprecated way to pass events to reverse proxy endpoints


kevinbader:
We support two ways of passing an event to a non-HTTP reverse proxy endpoint:
- Passing a CloudEvent and include a `target_partition` field in the event's `rig` extension.
- Passing a CloudEvent in an `event` field and the target partition next to it in a `partition` field.

Support for the latter should be dropped with the 3.0 release.

PR #335, opened 2020-10-29
334 fix endpoint validation

### LINKED ISSUE #334: Leaving out `path` in endpoint declaration let RIG fail on startup
## Summary and context

> Rewriting URLs in API Gateway only works when using regex catch groups, so we need to use `path_regex` instead of `path` when defining the path. Leaving out `path` declaration leads to a validation error on startup.

## How to reproduce

**When I run RIG with this endpoint:**

 ```
{
   "id": "get-entry",
   "path_regex": "/entries/(([\\d\\w])*(-([\\d\\w])*))",
   "path_replacement": "/api/entries/\\1",
   "method": "GET",
   "secured": false
}
```

**..I see the following error and/or log output:**

> module=RigInboundGateway.ApiProxy.Validations [error] Wrong reverse proxy configuration: [{"service-endpoint/get-entry", [{:error, "path", :by, "must be string"}, {:error, "path", :length, "must have a length of at least 1"}]}]

**..but I really expected this:**

> Normal startup because `path_regex` is set instead of `path`. This is also used in your tests and here [Forwarding Requests](https://accenture.github.io/reactive-interaction-gateway/docs/api-gateway.html#url-rewriting) 
OR 
> Allow dynamic parameters introduced by `{dynamic_paramter}` to be used in `path_replacement`, like
```
{
   "id": "get-entry",
   "path": "/entries/{entry_id}",
   "path_replacement": "/api/entries/{entry_id}",
   "method": "GET",
   "secured": false
}
```

 - Latest docker image `accenture/reactive-interaction-gateway:latest`
 - Reactive Interaction Gateway 3.0.0-alpha.1 [rig@127.0.0.1, ERTS 10.7.2.3, OTP 22]

### PR DESCRIPTION
**Pull request by mmacai**
## Description

Closes #334 

- fixed proxy endpoint validation to not require `path` when `path_regex` is set

## What to look out for

Dear reviewer, I want you to

You can try to run RIG with this (on this branch vs master branch):

```
PROXY_CONFIG_FILE='[{"id":"my-service","name":"my-service","version_data":{"default":{"endpoints":[{"id":"my-endpoint","method":"GET","path":"/","path_replacement":"/different-endpoint"},{"id":"my-transformed-endpoint","method":"GET","path_regex":"/foo/([^/]+)/bar/([^/]+)","path_replacement":"/bar/\\1/foo/\\2"}]}},"proxy":{"use_env":true,"target_url":"API_HOST","port":7070}}]' \
mix phx.server
```

## Process

Have fun!

### DIFF
--- a/lib/rig_api/v3/apis_test.exs
+++ b/lib/rig_api/v3/apis_test.exs
@@ -367,10 +367,7 @@ defmodule RigApi.V3.APIsTest do
                "invalid-config/" => [
                  %{"id" => "must be string"},
                  %{"id" => "must have a length of at least 1"},
-                 %{"path" => "must be string"},
-                 %{"path" => "must have a length of at least 1"},
-                 %{"path_regex" => "must be string"},
-                 %{"path_regex" => "must have a length of at least 1"},
+                 %{"path, path_regex" => "Either path or path_regex must be set"},
                  %{"method" => "must be string"},
                  %{"method" => "must have a length of at least 1"}
                ]
--- a/lib/rig_api/v3/apis_test.exs
+++ b/lib/rig_api/v3/apis_test.exs
@@ -357,7 +357,7 @@ defmodule RigApi.V3.APIsTest do
              }
     end
 
-    test "should return 400 when 'endpoint' doesn't have required properties 'id', 'method' and 'path'" do
+    test "should return 400 when 'endpoint' doesn't have required properties 'id', 'method' and 'path' or 'path_regex'" do
       endpoints = [%{}]
       new_api = ProxyConfig.create_proxy_config(@invalid_config_id, endpoints)
       conn = build_conn() |> put("/v3/apis/#{@invalid_config_id}", new_api)
@@ -369,6 +369,8 @@ defmodule RigApi.V3.APIsTest do
                  %{"id" => "must have a length of at least 1"},
                  %{"path" => "must be string"},
                  %{"path" => "must have a length of at least 1"},
+                 %{"path_regex" => "must be string"},
+                 %{"path_regex" => "must have a length of at least 1"},
                  %{"method" => "must be string"},
                  %{"method" => "must have a length of at least 1"}
                ]
--- a/lib/rig_api/v3/apis_test.exs
+++ b/lib/rig_api/v3/apis_test.exs
@@ -374,6 +374,67 @@ defmodule RigApi.V3.APIsTest do
              }
     end
 
+    test "should return 400 when 'endpoint' has 'path' and 'path_regex' set at the same time" do
+      endpoints = [
+        %{
+          "id" => @invalid_config_id <> "1",
+          "method" => "GET",
+          "path" => "/",
+          "path_regex" => "/"
+        }
+      ]
+
+      new_api = ProxyConfig.create_proxy_config(@invalid_config_id, endpoints)
+      conn = build_conn() |> put("/v3/apis/#{@invalid_config_id}", new_api)
+      response = json_response(conn, 400)
+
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: lib/rig_api/v3/apis_test.exs [hunk 1]
```diff
@@ -369,6 +369,8 @@ defmodule RigApi.V3.APIsTest do
                  %{"id" => "must have a length of at least 1"},
                  %{"path" => "must be string"},
                  %{"path" => "must have a length of at least 1"},
+                 %{"path_regex" => "must be string"},
+                 %{"path_regex" => "must have a length of at least 1"},
```
kevinbader, line 373: not sure, but this looks like both `path` and `path_regex` are required to be there and non-empty, while in fact exactly one of them must be present.

lib/rig_inbound_gateway/api_proxy/validations.ex
```diff
@@ -88,6 +88,16 @@ defmodule RigInboundGateway.ApiProxy.Validations do
 
   # ---
 
+  @spec validate_endpoint_path(Api.endpoint()) :: error_list_t()
+  def validate_endpoint_path(endpoint) do
+    errors = validate_string(endpoint, "path") ++ validate_string(endpoint, "path_regex")
```
kevinbader, line 93: what if both are valid? Indeed, I'm not sure what should happen in case _both_ properties are set. If you are, please add a test that documents/explains that behavior.

kevinbader (changes requested):

lib/rig_inbound_gateway/api_proxy/validations.ex
mmacai: No idea, what was your intention when adding `path_regex`?

mmacai:
@kevinbader Updated it in a way that it should be more clear and also catch the use-case when multiple path types are set.

lib/rig_inbound_gateway/api_proxy/validations.ex
kevinbader: The idea was to [keep `path` simple](https://github.com/Accenture/reactive-interaction-gateway/blob/master/lib/rig_inbound_gateway/api_proxy/api.ex#L15) and not introduce a breaking change there, but [add `path_regex` as a more powerful alternative](https://github.com/Accenture/reactive-interaction-gateway/blob/master/lib/rig_inbound_gateway/api_proxy/api.ex#L19).

Turns out, the way I've implemented it [allows to use both at the same time, and it'll just pick whatever matches first](https://github.com/Accenture/reactive-interaction-gateway/blob/master/lib/rig_inbound_gateway/api_proxy/api.ex#L100). Not sure that was intended, I don't quite recall.

To sum up, perhaps it's best to simply allow both be present and mention that if that's the case, only one of them needs to match. Additionally, we should consider deprecating `path` to remove it with 3.0.

mmacai: Hmm, that sounds a bit unpredictable and dangerous to me. Would rather prefer either one or another as it anyway probably doesn't make sense to use both at the same time.

[Here](https://github.com/Accenture/reactive-interaction-gateway/issues/278) we can get rid of everything deprecated and path as well if we want.

kevinbader: > Hmm, that sounds a bit unpredictable and dangerous to me. Would rather prefer either one or another as it anyway probably doesn't make sense to use both at the same time.

Unfortunately, changing this behavior now would be breaking. Then again: I don't expect anyone using both directives, since it doesn't make sense, like you said.

kevinbader (approved):

Issue #381, opened 2021-05-31
Automatic subscription via JWT claims doesn't support array of strings as values


patrickmedina:
## Summary and context

> Automatic subscription only supports strings as claims value. When passing claims with array of strings, the automatic subscription doesn't work at all. The most common use case for this is the "Roles" claim which is typically provided as an array of strings. Though this subscriptions can be added later after initiating the websocket/sse connection, it would be beneficial to have this as part of the automatic subscription via JWT claims during connection.

## How to reproduce

**When I run RIG with these settings:**

> _JWT_
```
{
    "iss": "Online JWT Builder",
    "iat": 1622458324,
    "exp": 1653994324,
    "aud": "www.example.com",
    "sub": "jrocket@example.com",
    "GivenName": "Johnny",
    "Surname": "Rocket",
    "Email": "jrocket@example.com",
    "Roles": [
        "Admin",
        "Tester"
    ]
}
``` 
> _Example Event_
```
{"data":{"clientId":"Personnel30","name":"User1","position":{"x":5170,"y":-2480,"z":370},"timestamp":"12304967032","userType":"Admin"},"id":"58bc0707-a9d7-41f8-91eb-12db371ff272","source":"test","specversion":"0.2","time":"2021-05-04T02:58:16Z","type":"position"}
```
> _Extractor json_
```
{
        "position": {
                "userType": {
                        "stable_field_index": 1,
                        "event": {
                                "json_pointer": "/data/userType"
                        },
                        "jwt": {
                                "json_pointer": "/Roles"
                        }
                }
        }
}
```

**..and set up clients and services like this:**

> _Websocket connection_
```
[...]

PR #386, opened 2021-07-13
Added support for array of values in JWT claims in auto-subscription

patrickmedina:
## Description

> Attempting to fix issue #381 @kevinbader 
>
> This change checks for the claims value type. If the JWT claim value from the json_pointer is an array of strings(E.g. {Roles: ["Administrator", "User"]}), it will construct a subscription with multiple entries(E.g. {...,"oneOf":[{"userType":"Administrator"},{"userType":"User"}]}],...}).

### DIFF
--- a/lib/rig/subscriptions/parser/jwt.ex
+++ b/lib/rig/subscriptions/parser/jwt.ex
@@ -52,6 +52,13 @@ defmodule RIG.Subscriptions.Parser.JWT do
     end
     |> Result.filter_and_unwrap()
     |> Enum.into(%{})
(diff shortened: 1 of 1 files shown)


kevinbader:
Nice work! Even though JSON pointer for some reason has no way of pointing to any value within an array rather than the array itself, I think this is a good, pragmatic solution. What left to do before merging this:
* a test for this to /lib/rig/subscriptions/parser/jwt_test.exs
* an entry to the CHANGELOG.md file

Could you please add that to this PR?

